PR #2, opened 2026-07-16
Revive dead sessions with v

YoanWai:
## What

`v` on a dead session recreates its tmux session under the same id, keeping name, group, and ordering. Per-tool `revive_command` decides what launches: the defaults resume the agent's last conversation (`claude --continue`, `opencode --continue`), falling back to the plain `command` when unset. Hook-backed tools get their settings flag and status-file environment re-injected, and the row returns to the tool's default status until the next poll.

Guards: revive refuses sessions that are still running and working directories that no longer exist.

## Verification

- `go build`, `go vet`, `gofmt`, `go test ./...` green; new tests cover the revive round-trip against a real tmux server, the still-running guard, and the missing-directory guard.
- `claude --continue` resumption verified live: a Haiku conversation stored a codeword, a second `--continue` invocation in the same directory recalled it.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -1,0 +1,25 @@
+name: CI
+
+on:
+  push:
+    branches: [main]
+  pull_request:
+
+jobs:
+  test:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v4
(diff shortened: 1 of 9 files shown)


YoanWai:
recheck

PR #24, opened 2026-07-18
Revive resumes each session's own conversation

YoanWai:
## Problem

Reviving a dead session relaunched the tool with its cwd-scoped resume command (`claude --continue`, `codex resume --last`, `grok --continue`, `opencode --continue`). Those resume the working directory's *most recent* conversation. When sessions share a directory, every revive resumed the same one, so all but the newest came back as the wrong conversation.

## Fix

Each session records its agent conversation id; revive resumes that exact id via a per-tool `resume_by_id_command` (`{id}` placeholder).

- **claude, grok** launch with `session_id_flag = "--session-id"`, so the id is chosen up front and revive is fully deterministic.
- **codex, opencode** mint their own id. The poller captures it from their session store (`~/.codex/sessions`, opencode `storage/session`) once the pane is live, matched by working directory and launch time, and deduped so two sessions started in one directory never claim the same conversation.

A new `agent_session_id` column (with migration) stores the id. Existing configs inherit the new tool defaults through the tool-default backfill.

## Verification

- `claude --session-id <uuid>` then `claude --resume <uuid>` round-trips against the real CLI.
- New tests: capture for codex and opencode (cwd match, launch-time cutoff, claimed dedup), store round-trip, config defaults and backfill.
- Full suite passes; `go vet` clean.


YoanWai:
Rebased onto main. Also fixes `tmux_test.go:83` which called the new 6-arg `Create` with 4 args — a build break that landed on main when #22 and #23 merged separately. Addressed two review findings: capture now resolves symlinks before comparing cwd (macOS /tmp vs /private/tmp).

PR #32, opened 2026-07-19
fix: resolve review repo when session cwd is an umbrella of repos

YoanWai:
## Problem

Review showed **"not a git repository"** for any session whose working directory is a container folder holding several repos, not a repo itself. Real cases from live sessions:

- `monorepo` (9 sessions) → 20 nested repos (`servers/main-node`, `apps/`, `libs/`, …)
- `MySpaze` (1) → `myspaze-app`, `myspaze-api`, `myspaze-admin`
- `projects` (1) → every project underneath

The diff loader ran git in the cwd, which is genuinely not a repo, so review was dead for all of these.

## Fix

When the cwd is not a repo, `ResolveRepos` discovers the nested repos (depth ≤ 3, skips `node_modules` / `.worktrees` / build output) and ranks them **dirty working tree first, then most recent commit**, so review opens on the repo the agent is most likely editing.

- `r` cycles between repos in the umbrella.
- Header pill and footer show the selected repo and `i/n`.
- A cwd that is itself a repo behaves exactly as before (no pill, no cycle).

Verified ranking on the real dirs: monorepo → `main-node`, MySpaze → `myspaze-app`, projects → `agent-manager`.

## Tests

- `git`: `ResolveRepos` single / umbrella-ranks-dirty-first / no-repo.
- `ui`: umbrella session opens on the active repo, header names it, `r` cycles and wraps — driven through the real `Update` loop.

Full suite, `go vet`, `gofmt` all clean.

### DIFF
--- a/internal/git/git.go
+++ b/internal/git/git.go
@@ -9,6 +9,7 @@ import (
 	"os"
 	"os/exec"
 	"path/filepath"
+	"sort"
 	"strconv"
 	"strings"
 )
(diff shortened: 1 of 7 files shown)


YoanWai:
### Code review

Found 2 issues (both fixed in dc3ebc252b28a3562b10f5aa53558e193fd4617d):

1. `repoIdx` was treated as a stable repo identity, but `ResolveRepos` re-ranks the repo list (dirty first, then recency) on every load, so between polls the index could silently point at a different repo. Fixed by pinning the selection to the repo root path and re-deriving the index each load.

https://github.com/YoanWai/agent-manager/blob/9de843b2bae5791a48bf81dd8602c508ab5ef143/internal/ui/diffview.go#L98-L108

2. Comments and reviewed marks were keyed by session only, so cycling to a sibling repo with a same-named file (e.g. `README.md`) inherited the wrong marks and could send a comment against the wrong repo. Fixed by keying review state per session and repo (`reviewKey()`); a stale probe against the previous repo is now ignored.

https://github.com/YoanWai/agent-manager/blob/9de843b2bae5791a48bf81dd8602c508ab5ef143/internal/ui/diffview.go#L44-L45

Regression tests added: repo selection survives re-rank, and reviewed marks stay isolated per repo. Full suite, vet, gofmt clean.

PR #33, opened 2026-07-19
feat: attach clipboard images in the quick prompt for any provider

YoanWai:
## What

Adds `Ctrl+V` in the quick-prompt bar to attach a clipboard image. The image is written to a temp file and its local path is inserted into the prompt, so the agent in the target session opens it on submit.

## Why provider-agnostic

Agent-manager spawns agents in local tmux on the same host, so any file the TUI writes is readable by the agent. Passing the image as a local path in the prompt text means every provider (claude, codex, grok, opencode, and future tools) reads it with no per-tool config.

## How

- New `internal/clipboard` package: `ReadImage()` reads the OS clipboard image, `SaveToTemp()` writes it under a shared pastes dir.
  - macOS: `osascript` coerces the clipboard to PNG.
  - Linux: `wl-paste` (Wayland) or `xclip` (X11), whichever is installed.
  - `goos` / `lookPath` / `runCmd` are injectable so tests drive every branch without a real clipboard.
- `handleQuickKey` intercepts `ctrl+v` and calls `attachQuickImage`, which surfaces a clear error on an empty clipboard, a missing Linux tool, or an unsupported OS.
- Help modal documents the keybind.

## Verification

- `go test ./...` passes.
- macOS clipboard read verified live end-to-end: a PNG on the clipboard is written byte-identical; a text clipboard errors and returns `ErrNoImage` with no file created.

## Scope

One image per paste, macOS + Linux. No drag-drop, Windows, or inline thumbnail.

### DIFF
--- a/docs/superpowers/specs/2026-07-19-paste-images-quick-prompt-design.md
+++ b/docs/superpowers/specs/2026-07-19-paste-images-quick-prompt-design.md
@@ -1,0 +1,41 @@
+# Paste images in the quick prompt
+
+## Goal
+
+Let a user attach a clipboard image while the quick-prompt bar is open, and have the image reach whatever agent runs in the target session, regardless of provider (claude, codex, grok, opencode, or any future tool).
+
+## Approach
+
+Agent-manager always spawns agents in local tmux on the same host, so any file the TUI writes is readable by the agent. On a `Ctrl+V` keypress in the quick bar, read the OS clipboard for image data, write it to a temp file, and insert that file's absolute path into the prompt text. On submit the existing `SendText` delivers the whole prompt (path included) via `send-keys`; the agent opens the local path.
+
+Nothing is tool-specific: every agent reads a local image path from prompt text. This is the only provider-agnostic mechanism and needs no per-tool config.
+
+## Components
+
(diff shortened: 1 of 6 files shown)


YoanWai:
### Code review

Reviewed with 3 parallel agents (CLAUDE.md compliance, bugs, git history). Two real issues found and fixed in this branch:

1. `ctrl+v` shadowed the textarea's built-in text paste, so a text-only clipboard failed with "no image in clipboard" and dropped the paste. Now `attachQuickImage` returns whether it handled the key; an empty clipboard falls through to the textarea's own paste.

https://github.com/YoanWai/agent-manager/blob/d51b1f71aab132647511c3278418056225465f76/internal/ui/keys.go#L734-L740

2. The inserted image path could scroll off the height-capped quick bar without following the cursor, since `InsertString` skips `repositionView`. A no-op `Update` after the insert now scrolls it into view.

https://github.com/YoanWai/agent-manager/blob/d51b1f71aab132647511c3278418056225465f76/internal/ui/keys.go#L678-L686

Both covered by new tests. No other issues in the clipboard readers, temp-file lifecycle, or error handling.

PR #37, opened 2026-07-20
fix: stop review listing entries it cannot diff

YoanWai:
## Problem

Review listed 20 files for a session where only 2 had reviewable changes. The other 18 rendered `+0 −0`:

- `.worktrees/TICKET-1/` and three siblings are **directories**. `git ls-files --others` reports a nested repository as a directory, because git will not descend into another repo. There was nothing to diff and the rows could not be opened.
- Untracked screenshots are **binary**, so no line counts exist.
- Untracked text files showed `+0 −0` too, because `git diff --numstat` covers only tracked changes.

The header total was wrong for the same reason: "20 files · +8 −4" counted only the 2 tracked files.

## Fix

- Skip untracked directory entries; a nested repo's changes belong to its own review.
- Count lines for files `numstat` does not cover, streamed from the working file at set-build time. Counting from raw bytes rather than from the built diff keeps the number true for files past the 10k-line display cap and the 1 MB byte cap, and doing it before the eager-load cap keeps the header total stable instead of climbing as you scroll.
- `numstat` stays authoritative wherever it had an entry, so a truncated tracked file keeps its real count.
- Render `binary` where a line count is meaningless, and `?` for a file too large to count, with the header marking the total incomplete rather than silently under-reporting.
- A working file that cannot be read marks that row and leaves the rest of the review open.

Header totals correct themselves once the per-file stats are right.

## Tests

- A nested repository is omitted while untracked files are still listed.
- An untracked file reports its true line count past both the line cap and the byte cap; an empty one stays at `+0 −0`.
- A truncated tracked file keeps its `numstat` value.
- The header total is correct before any lazy loading, and marks itself incomplete when a file is uncounted.
- An unreadable untracked file marks its own row without aborting the set.
- A binary file is labelled `binary` on its row.

Verified against `monorepo/servers/main-node`, where the four `.worktrees/TICKET-*` rows are gone.

Full suite, `go vet`, `gofmt` clean.

### DIFF
--- a/docs/superpowers/plans/2026-07-20-diff-noise.md
+++ b/docs/superpowers/plans/2026-07-20-diff-noise.md
@@ -396,10 +396,10 @@ Expected: the command prints the new pull request URL
 
 ## Self-Review
 
-**Spec coverage.** The spec's Phase 1 lists three defects. Untracked directories map to Task 1, missing line counts to Task 2, binary labelling to Task 3. The spec's claim that header totals correct themselves is verified in Task 3 Step 5 and stated in the PR body. The spec's Phase 1 test list maps to the tests in Tasks 1 through 3, plus the real-repository check in Task 4.
+**Spec coverage.** The spec's Phase 1 lists three defects. Untracked directories map to Task 1, missing line counts to Task 2, binary labelling to Task 3. The spec's claim about header totals is verified by `TestHeaderTotalStableWithoutLazyLoad` in `internal/diff/diff_test.go`, which asserts every listed file carries its stat straight out of `BuildSet`, past the eager-load cap and with no lazy load, so the total the header sums is correct on the first paint and does not drift as the user scrolls. The spec's Phase 1 test list maps to the tests in Tasks 1 through 3, plus the real-repository check in Task 4.
 
 **Placeholders.** Every code step carries the actual code. No TBD, no "handle edge cases", no "similar to Task N".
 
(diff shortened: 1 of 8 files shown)


YoanWai:
### Final review

Two independent reviews of the branch. Both found the same defect, now fixed in 6265dbc:

- **Binary label missed tracked binaries past the eager-load cap.** The row keyed off `fd.Binary`, set only by `loadFile`, which runs for the first 200 files or lazily under the cursor. A tracked binary at index >= 200 rendered `+0 −0` until you moved onto it — the exact bug this PR set out to kill. `git.FileStat.Binary` already carried the answer from numstat's `-`/`-` columns and was never read. Now `fd.Binary || fd.Stat.Binary`, with a regression test that pins the row as unloaded.
- **Counting ran in every scope.** `countUnknownStat` reads the working tree, but the new side is `:0` for staged and `HEAD` for last-commit/vs-base. Unreachable in practice (numstat covers everything name-status lists, verified for unmerged paths, submodule bumps, typechange and `-M` renames), so it was speculative code guarding a silent-wrong-data hazard. Now gated on `file.Status == git.Untracked`, which also skips the work in three of four scopes.
- **Counting budget 64 MiB -> 8 MiB.** `Fingerprint` hashes mtimes, so an agent appending to a large untracked log re-scanned it every poll. Measured 5000 untracked files at 183 ms of counting; pathological files now degrade to `?` instead.
- Strengthened `TestTruncatedTrackedFileKeepsNumstat`, which previously passed against pre-fix code, and recorded the shipped implementation in the plan doc.

Known follow-up, deliberately not in this PR: when a nested-repo directory is the only untracked entry, review shows `✓ no changes (uncommitted)`, which is inaccurate. Fixing it means plumbing a skipped-count through `ChangedFiles` into `diff.Set`.

PR #41, opened 2026-07-20
feat: confirm before archive/restore sessions

YoanWai:
Archive/restore now uses same bottom-bar confirmation pattern as delete.

- `confirmTarget.action` field dispatches archive, restore, or delete in `handleConfirmKey`
- `archiveSelected()`/`restoreSelected()` build confirm target + switch to confirm mode instead of calling store directly
- `archivalSnapshot()` extracted from old inline flow, called during confirm for archive
- Test call sites updated to go through confirm flow


YoanWai:
### Code review

Found 2 issues:

1. A failed `SetSnapshot` no longer aborts the archive. `archivalSnapshot()` returns `void` and only stashes the error in `m.err`; the caller proceeds to `SetArchived`/`SetGroupArchived` unconditionally, so a session can end up archived with no pane snapshot. This reverts the guard added in #36 (`9ba871f`), where `snapshotForArchive` returned an error and `setSelectedArchived` bailed before flipping the DB flag. It also contradicts the new doc comment promising the snapshot "survives when the tmux window is later killed".

https://github.com/YoanWai/agent-manager/blob/a20ea51ac1a0b227d821284914a5e5297b671871/internal/ui/keys.go#L545-L556

https://github.com/YoanWai/agent-manager/blob/a20ea51ac1a0b227d821284914a5e5297b671871/internal/ui/keys.go#L455-L473

2. A successful archive/restore no longer clears `m.err`. The removed `setSelectedArchived` did `m.err = ""` after a successful `SetArchived`/`SetGroupArchived`; neither new branch does, and `handleConfirmKey` has no `m.err = ""` anywhere. A stale error banner from an earlier failure keeps showing after a subsequent successful archive or restore.

https://github.com/YoanWai/agent-manager/blob/a20ea51ac1a0b227d821284914a5e5297b671871/internal/ui/keys.go#L553-L566

<sub>- If this code review was useful, please react with 👍. Otherwise, react with 👎.</sub>

YoanWai:
Both addressed in ce0fd20:

1. `archivalSnapshot` returns `error` again; `handleConfirmKey` bails before flipping the archived flag, restoring the 9ba871f guarantee.
2. Archive and restore now clear `m.err` on success. Covered by `TestArchiveRestoreClearStaleError`, verified failing without the fix.

Also folded in the CodeRabbit note: `actionArchive`/`actionRestore`/`actionDelete` constants, shared `applyConfirmedArchived` helper, and an unknown action reports an error instead of falling through to deletion.

PR #48, opened 2026-07-20
feat(ui): mouse-resizable sessions/sidebar split

### LINKED ISSUE #42: Mouse-resizable split between sessions and sidebar
Let the user drag the divider between the Sessions column and the right sidebar to set the split, entered through a held key so normal terminal text selection keeps working.

## Today

The split is a fixed percentage. `splitWidths()` in `internal/ui/view.go:57` computes `leftWidth = m.width * 34 / 100`, clamped to a minimum of 30, and returns `m.width - leftWidth` for the right side. It is a pure function of `m.width`, recomputed on every `View()`, and every consumer reads from it, so the split has exactly one source of truth to change.

The program runs with `tea.NewProgram(model, tea.WithAltScreen())` at `main.go:103`. Mouse reporting is off and `Update` has no `tea.MouseMsg` case.

## Design

Hold a modifier key to enter resize mode, drag the divider, release to commit. Gating on a held key keeps mouse reporting off during normal use, so selecting and copying text from the preview pane behaves exactly as it does now.

1. Enable mouse reporting only while resize mode is active, and turn it back off on exit.
2. Add a `splitRatio` field plus a drag state flag to the model; handle press, motion, and release on the divider column.
3. Commit the tmux resize on release. `previewPaneWidth()` feeds `tmux.Resize` for every non-archived session (`internal/ui/model.go:360`), so resizing on each motion event would fire a full sweep per frame. Redraw live during the drag, resize panes once at the end.
4. Persist the ratio in the `settings` table, following `collapsed_groups` (`internal/ui/model.go:197`) and `diff_layout` (`internal/ui/keys.go:943`).
5. Keep the existing minimum of 30 columns for the left side, and add the matching floor for the right.

## Scope

Horizontal split only. The vertical split inside the left column (sessions list above the Computer stats block, `internal/ui/view.go:42-47`) sizes itself from the measured height of the stats content, so it belongs in a separate change.

YoanWai:
## Summary
- Gate mouse reporting behind `|` so normal terminal text selection stays intact.
- Drag the sessions/sidebar divider; redraw live; on release persist `split_ratio` and resize tmux panes once.
- Esc/`|` cancels mid-drag and restores the previous ratio.

`|` is a **toggle** (not a held modifier): bubbletea cannot arm mouse on a held key alone. Toggle-in / release-or-esc-out is the intentional terminal-safe design for #42.

Closes #42.

## Test plan
- [x] Unit: `env -u TMUX TMUX_TMPDIR=/tmp/amtest go test ./internal/ui/ -run 'Split|Clamp|ResizeMode|Drag|PressOff|NewLoads|BodyY|EnterResize' -count=1`
- [ ] Full: `env -u TMUX TMUX_TMPDIR=/tmp/amtest go test ./... -count=1`
- [ ] Manual: press `|`, drag divider, release; confirm split persists across restart
- [ ] Manual: confirm text selection still works when not in resize mode
- [ ] Manual: Esc mid-drag restores prior split; no stray mouse capture after exit

### DIFF
--- a/internal/ui/keys.go
+++ b/internal/ui/keys.go
@@ -24,7 +24,11 @@ func (m *Model) handleKey(msg tea.KeyMsg) (tea.Model, tea.Cmd) {
 		case "|", "esc":
 			return m.exitResizeMode(false)
 		case "q", "ctrl+c":
-			return m, tea.Quit
+			// Drop mouse reporting before quit so the terminal is not left
+			// in mouse mode if shutdown races the bubbletea teardown.
+			m.resizeMode = false
+			m.splitDragging = false
+			return m, tea.Sequence(tea.DisableMouse, tea.Quit)
 		default:
 			return m, nil
 		}
--- a/internal/ui/keys.go
+++ b/internal/ui/keys.go
@@ -17,6 +17,19 @@ import (
 )
 
 func (m *Model) handleKey(msg tea.KeyMsg) (tea.Model, tea.Cmd) {
+	// Resize mode owns the keyboard until the drag commits or the user
+	// cancels: other bindings would fight the mouse-gated session.
+	if m.resizeMode {
+		switch msg.String() {
+		case "|", "esc":
+			return m.exitResizeMode(false)
+		case "q", "ctrl+c":
+			return m, tea.Quit
+		default:
+			return m, nil
+		}
+	}
+
 	switch m.mode {
 	case modeForm:
 		return m.handleFormKey(msg)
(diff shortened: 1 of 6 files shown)


YoanWai:
Addressed the pending review:

1. **exit paths** — release now calls `exitResizeMode(true)` (same path as cancel with `false`).
2. **resize-only-on-release** — `TestDragResizesTmuxOnlyOnRelease` asserts motion leaves tmux width alone and release resizes once; plus enter-blocked / press-outside-body cases.
3. **body Y geometry** — `listChromeRows` + `listBodyHeight()` shared by `View` and `bodyYRange`, with `TestBodyYRangeMatchesListChrome`.
4. **`|` vs held key** — documented on the PR body as intentional toggle design.

PR #49, opened 2026-07-20
fix(ui): insert pasted image paths into the quick prompt

YoanWai:
## Summary
- Paste (`Ctrl+V`) now inserts the temp image path into the quick-prompt textarea as normal, editable text.
- Removes the attachment chip / `attachments` state that blocked deleting a pasted image.
- Submit still sends the path as part of the prompt so the agent can open the file.

## Test plan
- [x] `go test ./internal/ui/ -run 'TestQuickAttach|TestQuickCtrlV'`
- [ ] Open quick prompt, paste an image, confirm path appears inline
- [ ] Backspace / delete the path, confirm it is gone before submit
- [ ] Paste beside existing text, confirm space separation
- [ ] Submit with path still present, confirm agent receives the path

### DIFF
--- a/internal/ui/keys.go
+++ b/internal/ui/keys.go
@@ -845,7 +845,24 @@ func (m *Model) attachQuickImage() bool {
 		return true
 	}
 	// Space-pad so the path is a standalone, backspace-deletable token.
-	m.quick.input.InsertString(" " + path + " ")
+	token := " " + path + " "
+	// bubbles InsertString silently truncates at CharLimit; refuse when the
+	// full path would not fit so we never submit a broken partial path.
+	if limit := m.quick.input.CharLimit; limit > 0 {
+		if m.quick.input.Length()+len([]rune(token)) > limit {
+			_ = os.Remove(path)
+			m.err = "prompt is full"
+			return true
+		}
+	}
+	before := m.quick.input.Value()
+	m.quick.input.InsertString(token)
+	if !strings.Contains(m.quick.input.Value(), path) {
+		m.quick.input.SetValue(before)
+		_ = os.Remove(path)
+		m.err = "prompt is full"
+		return true
+	}
 	m.err = ""
 	return true
 }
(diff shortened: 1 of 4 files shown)


internal/ui/keys.go
```diff
@@ -813,23 +813,15 @@ func (m *Model) attachQuickImage() bool {
 		m.err = err.Error()
 		return true
 	}
-	m.quick.attachments = append(m.quick.attachments, path)
+	// Space-pad so the path is a standalone, backspace-deletable token.
+	m.quick.input.InsertString(" " + path + " ")
```
YoanWai: **[suggestion]** `textarea.InsertString` (bubbles v1.0.0) silently truncates or no-ops when `CharLimit` (2000) does not have room for the full padded path. `attachQuickImage` still returns true, clears `m.err`, and leaves the temp file on disk. Near the limit the user can submit a partial path that does not exist, or see no path at all with no error.

**Suggestion:** After `InsertString`, verify the full `path` is present in `m.quick.input.Value()`. If not, set `m.err` (e.g. prompt is full), remove the temp file, and leave prior input unchanged when possible.

internal/ui/ui_test.go
```diff
@@ -1115,54 +1115,60 @@ func TestQuickAttachImageRecordsAttachmentNotInput(t *testing.T) {
...
+	}
+	msg := m.quickMessage()
+	if !strings.HasPrefix(msg, "look at this ") {
+		t.Fatalf("compose should space-separate path, got %q", msg)
+	}
+	fields := strings.Fields(msg)
+	if len(fields) < 4 {
+		t.Fatalf("want text tokens plus path, got %q", msg)
+	}
+	path := fields[len(fields)-1]
```
YoanWai, line 1174: **[nit]** `TestQuickAttachImagePadsBesideExistingText` takes `fields[len(fields)-1]` as the path and only `os.Remove`s it. Unlike the sibling test, it never reads the file, so a wrong last token still passes.

**Suggestion:** `os.ReadFile(path)` (and assert contents) before remove, same as `TestQuickAttachImageInsertsPathIntoInput`.

YoanWai (commented):
Addressed: CharLimit guard + temp cleanup; pad test now reads the file. Follow-up commit 3b82007.

PR #69, opened 2026-07-22
fix(ui): re-arm mouse capture after tmux attach/detach

YoanWai:
## Why

The TUI scroll regression has returned four times. Root cause is in bubbletea, not our earlier fixes.

Bubbletea v1.3.10 `RestoreTerminal()` runs after every `tea.ExecProcess` (each tmux attach). It re-enables altscreen, bracketed paste, and focus reporting, but **not mouse mode**. So the first session attach/detach silently kills wheel capture, the wheel goes to the host terminal or outer tmux, and the manager scrolls off-screen. Startup `WithMouseCellMotion()` arms mouse once at boot, which is exactly why every prior fix worked until the first attach and then regressed.

## What

Re-issue `tea.EnableMouseCellMotion` on all four `attachDoneMsg` return paths: normal detach, attach error, review jump, and review-clear error. Wheel capture now survives every attach/detach.

`internal/ui/mouse_rearm_test.go` asserts detach emits the mouse-enable command, guarding against a fifth regression.

## Verification

- `go build ./...` clean
- new regression test passes
- full `internal/ui` suite passes

### DIFF
--- a/internal/ui/model.go
+++ b/internal/ui/model.go
@@ -569,7 +569,7 @@ func (m *Model) Update(msg tea.Msg) (tea.Model, tea.Cmd) {
 		if msg.err != nil {
 			m.err = msg.err.Error()
 			m.requestRefresh()
-			return m, nil
+			return m, tea.EnableMouseCellMotion
 		}
 		// Ctrl+R inside the session sets a marker before detaching; consume
 		// it here and jump straight to review for the session just attached.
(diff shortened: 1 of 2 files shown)


YoanWai:
### Code review

No issues found. Checked for bugs and CLAUDE.md compliance.

Verified: all four `attachDoneMsg` return paths re-arm mouse; `tea.Batch(cmd, tea.EnableMouseCellMotion)` safely tolerates a nil `cmd` from `openDiff`; re-enable cannot collide with resize mode (false on detach) and is idempotent.

https://github.com/YoanWai/agent-manager/blob/b2a96d176f3aad998342cbedd50db03173975d55/internal/ui/model.go#L554-L596

PR #97, opened 2026-07-26
Fix Codex quick prompt submission

YoanWai:
## What changed

- deliver quick prompts through tmux using bracketed-paste boundaries when the target application requests them
- add a regression test that verifies the pasted text is bracketed and the Enter key remains a separate input event

## Why

Codex applies paste-burst detection to unbracketed terminal input. The quick-prompt path pasted the message and immediately injected Enter, which Codex could consume as part of the paste. The prompt therefore remained in the composer until the user entered the session and pressed Enter manually.

## Impact

Quick prompts sent to Codex sessions now submit immediately. The change is compatible with other tools because tmux only emits bracket markers when the pane application has requested bracketed-paste mode.

## Validation

- reproduced the failure against Codex CLI 0.145.0
- verified bracketed paste submits the same prompt immediately
- `go test ./...`
- `git diff --check`

### DIFF
--- a/internal/tmux/tmux.go
+++ b/internal/tmux/tmux.go
@@ -233,7 +233,11 @@ func (d *Driver) pasteAndEnter(target, text string) error {
 	if _, err := d.run("load-buffer", "-b", buf, path); err != nil {
 		return err
 	}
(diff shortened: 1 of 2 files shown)


YoanWai:
Superseded by #99, which includes this Codex quick-prompt fix.

PR #120, opened 2026-07-30
Quick prompt: paste images inline at the cursor

### LINKED ISSUE #117: Quick prompt: paste images inline at the cursor
Pasted images in the quick prompt should land at the cursor, the way Claude Code does it, instead of collecting in front of the whole line.

## Today

Chips render in the textarea prompt prefix, so every chip sits left of all typed text regardless of where the caret was when ctrl+v fired (`quickInlineChipPrompt`, `internal/ui/view.go:237`; `syncQuickInlineChips`, `internal/ui/view.go:220`). Removal is positional in the same way: backspace only peels a chip when the caret is at the text start, and it always drops the last attachment (`popQuickAttachment`, `internal/ui/keys.go:1052`). On submit the paths are concatenated after the text (`quickMessage`, `internal/ui/keys.go:1077`), so an agent reading "compare this with that" gets both paths at the end with nothing marking which image belonged to which clause.

## Wanted

- ctrl+v inserts the chip inline at the caret, splitting the text there: `fix the header [Image 1] and the footer [Image 2]`.
- The pasting placeholder occupies that same caret slot while the clipboard read runs, and resolves into the chip in place.
- Chips are atomic in the text: arrow keys step over a chip as one unit, backspace/delete next to a chip removes that chip, and typing around it keeps the surrounding text intact.
- Chips reflow with the text across the bar's wrapped rows.
- On submit each chip becomes its path at the position it occupied, so the message preserves the order the user wrote.

## Acceptance

- Paste mid-sentence, keep typing after it: the chip stays where it was pasted and the text after it is untouched.
- Two chips in one prompt, delete the first one: the second stays put and its path still goes out.
- Submitted message interleaves paths with text in caret order.
- Multi-row prompt keeps chips aligned with their surrounding words.

YoanWai:
Closes #117.

A pasted image lands at the caret as an `[Image #N]` token that renders as a tinted chip. The token lives in the typed text, so it wraps and reflows with the words around it.

- ctrl+v inserts the chip at the cursor, spaced off the neighbouring words; the chip holds that spot while the clipboard read runs and settles in place when the path lands.
- The caret treats a chip as one unit: arrows step over it, backspace or delete beside it takes the whole chip, and the caret never rests inside one.
- A chip removed by any route (chip key, ctrl+u, word delete) releases its temp file; a failed or image-less read takes the chip back out along with the spacing the paste added.
- On submit each token becomes its path, so `compare [Image #1] with [Image #2]` reaches the agent as `compare /…/paste-a.png with /…/paste-b.png`.

## Verification

`go test ./...` green (isolated tmux socket). Live run in an isolated tmux server against a real clipboard image:

- pasted mid-sentence, kept typing: chip stayed put, surrounding text intact.
- left/right stepped over the chip in one move; delete beside it removed the chip and its temp file.
- two chips in one prompt, submitted to a session running `cat`: the pane printed `compare <path1> with <path2>`.
- long prompt wrapped to a second row with the chip whole and styled on that row.


YoanWai:
### Code review

Reviewed the diff directly (bug scan, prior review comments on the files it touches, and the comment guidance in `internal/ui/keys.go`). Two issues found, both fixed in 64be09e:

1. Char-limit truncation. `textarea.InsertString` silently truncates at `CharLimit` (2000), so a nearly full prompt could take half an `[Image #N]` token: the pattern would stop matching, the path would never be substituted, and the temp file behind it would be stranded. This is the same failure PR #49 guarded against before chips moved out of the text. ctrl+v now refuses with "prompt is full" before reserving anything.

2. Abandoned pastes leaked. Closing the bar with esc discards the prompt (reopening starts fresh), but the images it was holding stayed on disk. Esc now releases them.

Also added coverage for chip removal in a multi-line prompt, where the caret is restored through a full value rewrite.

Checked and found fine: the in-place filter in `dropQuickAttachment` has no live pointer readers, the padding given back on removal is only ever spacing the paste added, submit keeps the temp files the agent still needs, and a chip left pending blocks submit rather than sending a raw token.

PR #121, opened 2026-07-30
feat(ui): settings toggle to close the quick bar after sending

YoanWai:
## What

Settings gains an **after quick send** field. `stay open` keeps today's behavior: the quick bar stays armed so consecutive prompts flow without reopening it. `close` dismisses the bar once the prompt is delivered.

The choice persists as the `quick_prompt_close` setting and covers both quick-prompt paths: answering the selected session and spawning a new one from a group.

## Verified

- `go test ./...` green (`internal/ui` full suite, 86s).
- New tests: default stays open, stored `close` opts in, the bar closes after a send when enabled, and the settings field persists its toggle.
- Settings modal rendered in both states, showing the new row focused and cycling between `stay open` and `close`.

### DIFF
--- a/README.md
+++ b/README.md
@@ -76,7 +76,7 @@ Sessions run inside tmux (`am_*` namespace), so they survive the manager quittin
 | `space` | Quick prompt: answer the selected session, or spawn an agent in the selected group |
 | `ctrl+r` | Review the selected session's changes: full-screen whole-file diffs, line comments sent to the agent |
 | `f` | Fold / unfold group |
-| `s` | Settings (quick-spawn tool, theme, review layout) |
+| `s` | Settings (quick-spawn tool, theme, review layout, after quick send) |
 | `t` | Toggle archived view |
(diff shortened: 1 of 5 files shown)


YoanWai:
### Code review

Two findings, both fixed in the follow-up commit:

1. Stale doc comment: `submitQuick` still claimed the bar always stays active, which the new toggle makes conditional.
2. Dead assignment: the settings save synced `m.quick.closeAfterSend`, but `handleKey` routes every key to `handleQuickKey` while the bar is active, so settings can never be open at the same time. Removed rather than kept as speculative code.

https://github.com/YoanWai/agent-manager/blob/dc362d2a964d62b55cf5a09f4b657a89def3d02a/internal/ui/keys.go#L1180-L1184

Checked otherwise: field-index math still wraps over all four settings rows, `quick_prompt_close` follows the same store pattern as `diff_layout`, and consumers of `m.quick.active` in the list/split views see the same state the esc path already produced.

PR #125, opened 2026-07-30
feat(ui): report chip temperatures in the Computer block

YoanWai:
The rail's machine block ends with a temp row again: `cpu 61°C  gpu 55°C` where the hardware draws that line, `soc 67°C` where the chip reports itself whole. A machine with no readable sensor keeps the block exactly as it is today.

## What changed

- **Render.** `computerLines` writes the temp row under disk, in the same shape as the net row.
- **Naming coverage.** The classifier now reads AMD's `tctl`/`tccd`, `zenpower_tdie` as a CPU rather than a bare die, `x86_pkg_temp`, the `cpu-thermal` and `soc-thermal` zones an ARM board exposes, and `nouveau` beside the existing `amdgpu`/`nvidia` keys. Storage, wireless and battery sensors stay out, so a warm NVMe is never printed as a warm CPU.
- **Linux partial reads.** Linux returns the sensors it could read alongside a warning for the ones it could not. Those readings are now kept.
- **Cadence.** Each read builds an IOKit HID client on Apple Silicon and costs ~57ms, so a reading is reused for 5s. With gauges sampling every 4s that lands on a sensor read every 8s, measured from a running build: `21:34:39.520`, `:47.531`, `:55.553`, `21:35:03.531`, `:11.537`. Roughly 0.7% of one core, and the read already ran on every gauge sample before this change.

## Verification

- Live on Apple Silicon (macOS, `CGO_ENABLED` off): `temp soc 54°C` in the rail.
- The reading tracks real heat. Sampled every 3s across idle → load → idle: 52.5°C idle, rising to 78.3°C after 24s of full load, back to 55.1°C once the load stopped. The heatsink-side `tdev` sensors peak three seconds behind the dies, as they should.
- Hottest die is the figure shown. At idle the 16 dies span 52.6–60.9°C, so it reads ~4°C above their mean — worst case is what throttles.
- The Linux reader runs for real in `TestReadTempsFromSysfs` over a sysfs tree the test lays out: Intel `coretemp` beside an NVMe, AMD `k10temp` with a discrete GPU, a thermal-zone-only board, an SoC zone, and a kernel with no sensor files at all. It is linux-tagged, so CI exercises it on every push.
- `TestClassifyTemps` covers the names each platform reports, including the Apple Silicon dies read off this machine.
- `go test ./...` green.

Intel Macs are covered by code path only — gopsutil polls a fixed SMC key list (`TC0D`/`TC0H`/`TC0P`, `TG0D`/`TG0H`/`TG0P`) with no cgo, and the matchers cover those keys — but there was no Intel machine here to run it on. Same for WSL, which takes the Linux path and shows a reading only where that kernel exposes thermal zones.

### DIFF
--- a/internal/sysstat/sysstat.go
+++ b/internal/sysstat/sysstat.go
@@ -221,7 +221,7 @@ func cachedTemps() tempReading {
 func classifyTemps(read []sensors.TemperatureStat) tempReading {
 	var reading tempReading
 	for _, sensor := range read {
-		if sensor.Temperature <= 0 {
+		if !plausibleTemp(sensor.Temperature) {
 			continue
 		}
 		key := strings.ToLower(sensor.SensorKey)
@@ -247,6 +247,14 @@ func classifyTemps(read []sensors.TemperatureStat) tempReading {
 		reading.soc, reading.socOK = 0, false
 	}
 	return reading
+}
+
+// plausibleTemp keeps a sensor that is reporting silicon rather than its own
+// absence: an unpopulated SMC key reads 0, a detached probe can read NaN or
+// an infinity, and a confused driver can report thousands of degrees. The
+// comparison also rejects NaN, which no ordering would have caught.
+func plausibleTemp(celsius float64) bool {
+	return celsius > 0 && celsius <= 150
 }
 
 func isGPUSensor(key string) bool {
--- a/internal/sysstat/sysstat.go
+++ b/internal/sysstat/sysstat.go
@@ -6,6 +6,8 @@ import (
 	"runtime"
 	"strconv"
 	"strings"
+	"sync"
+	"time"
 
 	"github.com/shirou/gopsutil/v4/cpu"
 	"github.com/shirou/gopsutil/v4/disk"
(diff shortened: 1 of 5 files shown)


YoanWai:
### Code review

Found 2 issues, both fixed on this branch:

1. Readings after the first lost their color. The row joined its parts with an already-rendered separator, and `subtleStyle.Render("  ")` carries its own reset; nested inside the outer `valueStyle.Render(...)` that reset landed mid-string, so `gpu`/`soc` printed in the terminal default. Verified under a forced ANSI256 profile: `"\x1b[38;5;188mcpu 61°C\x1b[38;5;60m  \x1b[0mgpu 55°C\x1b[0m"`. Invisible on Apple Silicon, which reports one `soc` figure; it needed a machine reporting both cpu and gpu.

https://github.com/YoanWai/agent-manager/blob/2eef95f0e1d0dbb98f8a04ab5e1e6d8e4e6b6c1a/internal/ui/listview.go#L404-L420

Fixed in 8e12515: each reading renders itself and the separator is concatenated, matching the net row. `TestTemperatureReadingsEachKeepTheirColor` fails against the old line.

2. A NaN reading set the OK flag while storing 0, so the rail would have drawn `cpu 0°C`, and a positive infinity would have printed as `+Inf`. `sensor.Temperature <= 0` is false for NaN. Raised by CodeRabbit.

Fixed in 2eef95f: `plausibleTemp` keeps `0 < t <= 150`, which also drops the thousands-of-degrees readings a confused driver emits, with regression cases in `TestClassifyTemps`.

Checked and clean: cache locking around the shared `tempCache`, sensor classification precedence (`zenpower_tdie` reaches `isCPUSensor` before the die branch), the rail's gutter and label conventions, and consistency with 674ee32 and 7b37633. No past review comments on these files to reconcile against.

PR #137, opened 2026-07-30
docs: let the demo open the README instead of a badge row

YoanWai:
## What changed

The README opens on the demo. Title, one line of what it is, then the thing running.

The badge row is gone, and so is everything that fed it: `internal/badge`, `tools/badges`, `docs/badges/*.svg`, and the `Badges` workflow. 445 lines deleted, one scheduled job retired.

## Why

GitHub already prints stars, license, release and language in the sidebar beside the README, so a badge row mostly reprints what is on screen a few centimetres to the right. What it costs is the first screenful: six pills sat between the title and the only thing that actually sells a TUI, which is watching it work.

Removing them also removes a standing maintenance obligation. The chips needed a scheduled workflow to stay true, that workflow needed a token GitHub does not hand to Actions, and a clone count labelled `14d` would have frozen the moment the token went stale. None of that is worth carrying for numbers a reader can see anyway.

## How it was verified

Rendered the new header against `#ffffff` and `#0d1117`, GitHub's two page backgrounds, and screenshotted both: the demo now sits directly under the tagline in each.

Checked that nothing references what was deleted: no mention of `badges` or `internal/badge` survives in the README or under `.github/`, `go build ./...` and `go vet ./...` are clean.

- [x] `go vet ./...` passes
- [x] `go build ./...` passes
- [ ] Ran it against real sessions (no runtime code changed)


YoanWai:
Superseded by the stat card in #138: the stats stay, the badge row does not.

PR #143, opened 2026-07-30
feat(ui): focus mode - type into a session without leaving the manager

YoanWai:
## What

Pressing Enter on a session now focuses it in place: the list, rail and live preview stay on screen while the keyboard goes to the agent. Ctrl+Q returns. A still attaches full screen, and a settings row (`session keys`) swaps which key does which.

## How it works

- **Control-mode client** (`internal/tmux/control.go`): one persistent `tmux -C attach-session` per focused/selected session. tmux pushes `%output` the moment the pane paints; captures and commands ride the same pipe. 0.15ms per capture vs 7.3ms per exec fork (benchmarks included). Reply blocks are matched by tmux's `%begin`/`%end` tag so pane text that displays the protocol itself cannot desync the queue, commands time out after 2s (they run on the UI loop), and a failed write dequeues its waiter.
- **Event-driven preview** (`focuswatch.go`): the client follows the selection once the cursor settles, debounces paint bursts (25ms), and silences the slower poll/tick captures that repainted stale frames over fresh ones. A session whose client dies backs off 15s instead of costing a fork per poll pass.
- **Closed-window mouse** (`focussel.go`): mouse reporting is taken while focused, so drag / double / triple click select pane text only and copy to the system clipboard; clicks can never retarget the keyboard.
- **Scrolling** (`focusscroll.go`): wheel events forward as SGR reports to apps that own the mouse (agent CLIs on the alternate screen); plain panes scroll tmux history with the view pinned while scrolled back.
- **Caret**: drawn from tmux's reported cursor cell, blinking at 530ms, spliced in by display column so the agent's own colours survive around it.

## Performance

Fixed-work comparison against main, 20 sessions, isolated socket and fake HOME, 3 runs each (CPU-seconds per fixed work / RSS):

| | idle 15s | 300 cursor moves | RSS |
|---|---|---|---|
| main | 0.77 | 0.45 | 41MB |
| this branch | 0.57 | 0.31 | 33MB |

## Testing

- Full suite green on an isolated tmux socket, including live integration tests for the control client, pushed previews, cursor reporting, wheel forwarding into a mouse-tracking pane, and failure backoff.
- Protocol regression tests: reply-queue desync via protocol-lookalike pane text, command timeout, greeting handling, server-exit waiter wakeup.
- Concurrency audit of the control client and watcher (deadlock class: watcher blocked in `program.Send` while Update stops it) with a regression test that reproduces the hang against the old code.


YoanWai:
@coderabbitai All 12 findings addressed in a652c14:

- **control.go write-under-mutex**: writes moved out of `mu` behind a dedicated write lock that also pins queue order (`submit`).
- **control.go unbounded Close**: bounded wait, then kill.
- **clipboard CombinedOutput hang**: stdout/stderr left unwired (no inheritable pipes) plus a 3s context timeout.
- **double-send fallback**: forwarding now uses `Control.Send` via `attempt()` - fallback fires only when nothing went over the pipe.
- **sync queries on the wheel path**: the watcher pushes mouse ownership and history depth with every capture; `wheelFocus`/`scrollFocus` read the cached values, no tmux round trips in Update.
- **stopLocked stale client**: cleared immediately on stop.
- **status order**: errors now render above the focus/scrolled notices.
- **-H vs tmux 3.0**: README bumped to tmux 3.1+.
- Test fixes: syncBuf for -race, waitWritten in the error-block test, failure-stamp assertion for the backoff retry, pre-quick-bar height comparison.

Full suite green on an isolated socket, `internal/tmux` and `internal/clipboard` also pass under -race.

YoanWai:
@coderabbitai The 3.1+ line is intentional and factual: keystroke forwarding in focus mode uses `send-keys -H`, which tmux added in 3.1, so 3.0 can no longer run this feature. The installer and packages do not pin tmux versions anywhere; the README line is the single place the requirement is documented, and it now states the real floor.

PR #155, opened 2026-07-31
feat(ui): add a comfortable list density

YoanWai:
Settings gains a `list density` picker between theme and review layout.

- `compact` (default) keeps every entry on one line, as today.
- `comfortable` gives each session and group two lines: the name on top, its `status · tool · last activity` indented underneath, with the tree's open branches carried down through the second line. Groups show their status dots below their name. Long names stop being truncated on a narrow rail.

```
├─ ○ loose-session
│    idle · claude · 0s ago
╰─ ▾ Features
     ○ 2
   ├─ ○ density-modes
   │    idle · claude · 0s ago
   ╰─ ○ footer-padding
        idle · claude · 0s ago
```

The choice persists in the store (`list_density`) and is cached on the model, since every rail frame reads it. The rail also opens at 30% of the terminal instead of 34%, giving the preview the room the second line takes back.

Verified by rendering the real frame at 100x22 and 200x26 in both densities. Covered by `internal/ui/listdensity_test.go`: the settings toggle and its persistence, session and group stacking, guide continuity through the meta line, and a short rail keeping the selected entry whole. Full suite green.

### DIFF
--- a/README.md
+++ b/README.md
@@ -124,7 +124,7 @@ Agent sessions live on a private tmux server named `agentmgr`, so they never mix
 | `space` | Quick prompt: answer the selected session, or spawn an agent in the selected group |
 | `ctrl+r` | Review the selected session's changes: full-screen whole-file diffs, with `c` to comment a line and `C` to send the comments to the agent |
 | `F` | Fold / unfold every group |
-| `s` | Settings (quick-spawn tool, theme, review layout, after quick send) |
+| `s` | Settings (quick-spawn tool, theme, list density, review layout, after quick send) |
 | `\|` | Resize the split: `←→` nudge the divider, `enter` commits, `esc` cancels |
 | `t` | Toggle archived view |
 | `e` | Hide / show empty groups |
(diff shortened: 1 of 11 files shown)


YoanWai:
### Code review

Found 1 issue, fixed in b3800d2:

1. On a short rail, a two-line comfortable entry lost its meta line. `entryLines` prepended the "↑ N more" counter and appended the "↓ N more" one before the height trim, so with room for exactly one two-line entry (`listBodyHeight` floors at 3, leaving a 2-line entry budget) the trim cut the selected entry in half. It then read as a compact row that had simply lost its status, contradicting the invariant `lineWindow` documents: "scrolling by whole entries so an entry is never cut in half".

https://github.com/YoanWai/agent-manager/blob/de3866a60360a64b1b1832b005eacde2c996c2c2/internal/ui/listview.go#L140-L161

The counters now take only the room the entries leave, covered by `TestComfortableRowSurvivesShortRail`.

Also dropped a comment on `stackedRow` that restated what the function name already said.

YoanWai:
### Code review

No issues found. Checked the guide-continuity change (`treeGuideTrail` / `slotContinues`) for divergence between the name line and the meta line, the meta indent width against the head line's glyph column, the interaction with the pinned root row from #157, and the 30% ratio against the `minSplitSide` floor.

PR #160, opened 2026-08-01
fix(focus): keep pasted prompts in the composer instead of auto-submitting

YoanWai:
## Problem

Pasting into a focused session (Cmd+V) auto-submitted the prompt the moment the clipboard contained a newline. Focus mode forwarded the paste byte-by-byte via `send-keys -H`, so each newline reached the agent as a bare Enter keystroke. This hit every provider (Claude, Codex, Gemini) since the forwarding path is provider-agnostic.

## Fix

Bracketed-paste events from the terminal now route through `load-buffer` + `paste-buffer -p`: tmux re-wraps the text in bracketed-paste markers when the pane application requests them, and no Enter is sent. Agents keep the full multi-line text in their composer; the user submits when ready.

- `Driver.Paste` delivers text without submitting (shares the buffer path with `SendText`)
- `handleFocusKey` routes `Paste` key messages to it
- `focusKeyCommand` refuses paste messages so they can never ride the raw-bytes path

## Tests

- `TestPasteDeliversWithoutSubmitting`: real tmux pane with bracketed paste on receives exactly `ESC[200~text ESC[201~` and nothing after
- `TestFocusPasteKeepsPromptInComposer`: focus mode routes the paste as one block
- `TestFocusKeyCommand`: paste messages are rejected by the raw-key encoder

### DIFF
--- a/internal/tmux/tmux.go
+++ b/internal/tmux/tmux.go
@@ -219,11 +219,28 @@ func (d *Driver) SendText(id, text string) error {
 	return d.pasteAndEnter(sessionName(id), text)
 }
 
+// Paste delivers text into the session's pane without submitting it. The
+// focus path uses this for clipboard pastes: sending the bytes as raw
+// keystrokes would turn every newline into an Enter press and submit the
+// agent's prompt mid-paste.
+func (d *Driver) Paste(id, text string) error {
+	return d.paste(sessionName(id), text)
+}
+
 var pasteSeq atomic.Uint64
 
 // pasteAndEnter pastes text of any length into a pane and submits it.
-// tmux send-keys silently stops around 1024 bytes; load-buffer does not.
 func (d *Driver) pasteAndEnter(target, text string) error {
+	if err := d.paste(target, text); err != nil {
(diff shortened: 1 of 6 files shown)


YoanWai:
### Code review

No issues found. Checked for bugs and CLAUDE.md compliance.

Verified across four independent passes (diff bug scan, git history, prior PR review comments, in-code invariants):
- The `-p` bracketed-paste flag introduced in 84ffc2c survives the `paste()`/`pasteAndEnter()` split, so the Codex Enter-consumption fix holds on both paths.
- Prior review comments on #143 (tmux 3.0 `-H` non-ASCII limits, double-send fallback race) do not apply: the paste branch returns before the hex path and never retries.
- Design note, non-blocking: the paste branch bypasses the persistent control pipe from 23f537b/a652c14 and forks tmux directly. Pastes are rare and not latency-sensitive the way single keystrokes are, so the fork cost is an acceptable tradeoff.

<sub>- If this code review was useful, please react with 👍. Otherwise, react with 👎.</sub>

PR #172, opened 2026-08-01
fix(focus): scrolling dead after an agent finishes — keep cached pane state on focus entry

YoanWai:
## Bug

Focus-mode scrolling (not attached) sometimes went dead right when an agent finished its turn: the wheel did nothing until the agent painted again.

## Root cause

`focusSelected` wipes the cached pane state (`pane.mouse`, `pane.motion`, `pane.sgr`, `pane.history`) on every focus entry and relies on the focus watcher's next pushed capture to restore the real values. Two facts break that together:

1. The watcher follows the *selection*, so it is usually already streaming the session before focus is entered; `setFocus` is then a no-op and never re-captures.
2. A finished agent's pane is completely quiet (verified: an idle Claude pane produces zero repaints), so no `%output` event ever arrives to trigger the restoring capture.

Result: after entering (or re-entering) focus on a quiet pane, the wheel routes as "plain pane, no mouse, zero history", `scrollFocus` clamps every notch to offset 0, and scrolling is dead. While the agent is still working, pushes flow constantly and repair the state within milliseconds — which is why it only bit after a turn ended.

## Fix

Keep the cached state when the watcher is already serving this session: the cache is this pane's own, and any terminal-state change repaints the pane, which pushes a fresh capture. A fresh watcher still starts from a clean reset, so another session's flags can never route this pane's wheel.

## Verification

New test `TestFocusReentryKeepsPaneStateOnQuietPane` reproduces the exact flow against a real tmux server and control client: focus a mouse-tracking pane, leave, let it sit quiet, re-enter, wheel — asserts the state survives and the SGR wheel report reaches the pane. Failed before the fix, passes after. Full suite green.

### DIFF
--- a/internal/ui/focuskeys.go
+++ b/internal/ui/focuskeys.go
@@ -127,11 +127,11 @@ func (m *Model) focusSelected() (tea.Model, tea.Cmd) {
 	m.focusScroll = 0
 	// Pane state from a previously watched session must not route this
 	// one's wheel; a fresh watcher's first pushed capture reports the real
-	// values. When the watcher is already streaming this session the cache
-	// is this pane's own and stays: a quiet pane pushes nothing, so a
-	// reset here would leave the wheel routed as a plain pane with no
-	// history until the agent next paints.
-	if m.focus == nil || !m.focus.serving(sess.ID) {
+	// values. When the watcher is already streaming this session and the
+	// cache came from its own capture, it stays: a quiet pane pushes
+	// nothing, so a reset here would leave the wheel routed as a plain
+	// pane with no history until the agent next paints.
+	if m.focus == nil || !m.focus.serving(sess.ID) || m.pane.forID != sess.ID {
 		m.pane.mouse = false
 		m.pane.motion = false
 		m.pane.sgr = false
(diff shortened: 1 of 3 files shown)


YoanWai:
### Code review

Found 1 issue, fixed in c0aa2f4:

1. Race: the entry guard trusted `serving(sess.ID)` to mean the cached pane flags belong to this session, but the watcher registers its control client before the first capture lands (https://github.com/YoanWai/agent-manager/blob/c0aa2f4acd459c3d1bfc6d29875c948be340902a/internal/ui/focuswatch.go#L199-L211). Focusing a just-selected session in that window kept the previous session's flags routing the wheel. The cache is now stamped with the session whose capture wrote it and kept only for that session:

https://github.com/YoanWai/agent-manager/blob/c0aa2f4acd459c3d1bfc6d29875c948be340902a/internal/ui/focuskeys.go#L128-L141

Checked: bugs in the diff, AGENTS.md compliance, git history invariants (936d9db/988d55b/c859d6a/f0777c2 wheel-routing fixes), prior PR review comments, and in-code invariant comments. No other issues.

Issue #180, opened 2026-08-02
New group/session keyboard usage UX is a bit annoying


fruch:
### What happened

On new group or session window, when scrolling up and down with keyboard
it drop down directly to group selection, and change the root folder override anything you might have editing so far

also when opening a session to prompt, there not enough space for writing a decent prompt, and no handling of the overflow text as one might expect

 

### agent-manager version

0.17.0

### Operating system

Linux

### Which agent CLI

Claude Code

### Relevant config

```toml
I running it under Zilljh (tmux replacement), and Foot as the terminal.
```

### Screenshot or pane output

<img width="578" height="255" alt="Image" src="https://github.com/user-attachments/assets/f2b2faf0-1282-4b35-8fa7-a1961ced4270" />

YoanWai:
Thanks for the report! I'll fix it as soon as possible, hopefully tomorrow.

For the new-session issue, there's a workaround in the meantime: create an empty group with a path, then press SPACE on it to quick-launch an agent. Writing a long prompt there works properly.

I'll ping you here as soon as the fix lands in a release. Thanks again for trying it out and taking the time to report this!

YoanWai:
Thanks again for the report, both parts are now fixed and released in v0.17.1:

- The prompt field in the new-session card is a proper multi-line input now: long text wraps across rows, and long name/dir values scroll instead of clipping (#182).
- Arrow keys and tab only move between fields in the new-session and new-group cards; the group selection changes only with left/right, so scrolling can't silently switch the group or override the folder you were editing (#185).

You can update now, or wait for the update notice to appear in the app and update from there.

The quick-launch flow also still works if you prefer it: create an empty group with a path and press SPACE on it to launch an agent with a prompt.

YoanWai:
v0.17.2 is out, with the follow-ups to the same round:

- The notices modal keeps its border and key hints on a short terminal, and long messages wrap and mark where they were cut (#186).
- The worktree toggle reads `unavailable (not a git repo)` when the target directory cannot hold a worktree, and says why when pressed, instead of failing once the prompt is already typed (#187).

Your report is credited in the in-app message that ships with this release, so it shows up in everyone's messages card. Thanks again for taking the time to write it up.

PR #182, opened 2026-08-02
fix(form): scroll long name/dir inputs and wrap the prompt field

YoanWai:
Part 1 of #180.

The new-session card clipped every field at the card edge, so a long prompt (or a long directory) disappeared behind an ellipsis while typing.

- `name` and `dir` now set the textinput width to the card's value column, so long values scroll horizontally and the cursor stays visible.
- `prompt` is now a textarea (same widget as the quick prompt bar): text soft-wraps and the field grows up to 4 rows, with the viewport following the cursor beyond that.
- `formField` renders multi-line values by indenting continuation rows under the value column.
- Field widths re-sync on terminal resize.

Verified live in an isolated tmux run: long prompt wraps across rows, long dir shows its tail with the cursor. `go test ./...` green.

The keyboard-navigation half of #180 (arrow keys dropping into the group picker and resetting the dir override) comes in a second PR.

### DIFF
--- a/internal/ui/form.go
+++ b/internal/ui/form.go
@@ -13,8 +13,10 @@ import (
 	"github.com/YoanWai/agent-manager/internal/status"
 	"github.com/YoanWai/agent-manager/internal/store"
 	"github.com/YoanWai/agent-manager/internal/tmux"
+	"github.com/charmbracelet/bubbles/textarea"
 	"github.com/charmbracelet/bubbles/textinput"
 	tea "github.com/charmbracelet/bubbletea"
+	"github.com/charmbracelet/lipgloss"
 	"github.com/google/uuid"
 )
 
@@ -68,7 +70,7 @@ type groupOption struct {
 type form struct {
 	name         textinput.Model
 	dir          textinput.Model
-	prompt       textinput.Model
+	prompt       textarea.Model
 	dirAuto      bool
 	toolNames    []string
 	toolIndex    int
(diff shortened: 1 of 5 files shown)


YoanWai:
### Code review

Reviewed with three passes (bug scan against bubbles v1.0.0 sources, AGENTS.md/comment-guidance compliance, git history + CodeRabbit triage). Two real issues found, both fixed in 27d7e6e:

1. `textareaRows` undercounted a pasted line whose width is an exact multiple of the text width: bubbles renders a trailing empty wrap row, so `"1234567890\nx"` at width 10 takes 3 rows while the old formula returned 2, hiding the last line. Regression test added; also corrects the same estimate the quick bar inherited.

https://github.com/YoanWai/agent-manager/blob/27d7e6e11382aa880093957becda502f0d721ccf/internal/ui/view.go#L238-L245

2. Resizing the terminal with the form open updated `textinput.Width` without recomputing the scroll window (textinput only does that in Update/SetValue/SetCursor), leaving a stale window until the next keystroke. `syncFormFieldWidths` now nudges `SetCursor(Position())`.

https://github.com/YoanWai/agent-manager/blob/27d7e6e11382aa880093957becda502f0d721ccf/internal/ui/form.go#L160-L169

CodeRabbit's "allow newline insertion in the prompt textarea" is a false positive: enter-submits matches the quick bar, and multiline text arrives via bracketed paste which never hits the enter case.

Width math, narrow-terminal floors, enter interception, and the SetHeight-before-Update viewport invariant all verified clean.

fruch:
LOL, that was fast :)

PR #183, opened 2026-08-02
fix(notices): show every line of a message instead of cutting it

YoanWai:
Long message bodies were cut off in the messages modal: the content column was pinned at 62 characters, so a feed message wider than that ended in an ellipsis and the rest was unreadable.

The modal now sizes to its content. It widens to the longest body line across the listed messages, capped at the terminal width, and word-wraps any line that still does not fit. 62 stays as the floor, so short messages keep the size they had.

`TestNoticesLongBodyWrapsFully` covers a long feed body: the frame must widen past the old cap, the tail words must all appear, and no rendered line may exceed the terminal.

### DIFF
--- a/internal/ui/notices.go
+++ b/internal/ui/notices.go
@@ -410,11 +410,18 @@ func (m *Model) handleNoticesKey(msg tea.KeyMsg) (tea.Model, tea.Cmd) {
 }
 
 func (m *Model) viewNotices() string {
+	notices := m.activeNotices()
 	inner := noticeModalInner
+	for _, n := range notices {
+		for _, line := range n.body {
+			if w := lipgloss.Width(line); w > inner {
+				inner = w
+			}
+		}
+	}
 	if m.width >= 8 && inner > m.width-8 {
 		inner = m.width - 8
 	}
-	notices := m.activeNotices()
 	if len(notices) == 0 {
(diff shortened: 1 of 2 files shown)


YoanWai:
### Code review

Found 2 issues, both fixed in 5c4c13ec38a37b5875725c09b31329e772538bfc:

1. Wrapping made the modal taller than the terminal, and `clampFrame` then cut the bottom off: the tail of the body, the URL row, the key hints and the modal's own closing border all disappeared with nothing to say they had. Reproduced at 30x12 with a 140-character body. The body is now fitted to the rows the terminal can hold and marks the cut with an ellipsis.

https://github.com/YoanWai/agent-manager/blob/247284dc565988dec0e53cff94c38465f447e821/internal/ui/notices.go#L441-L448

2. The width cap only applied at 8 columns and up, so terminals of 1 to 7 columns (past the `m.width == 0` guard in `View`) sized the content column purely from the message text and drew wider than the screen. The cap now applies at every width with a floor of one column.

https://github.com/YoanWai/agent-manager/blob/247284dc565988dec0e53cff94c38465f447e821/internal/ui/notices.go#L421-L424

Covered by `TestNoticesShortTerminalKeepsFrameAndHint` and `TestNoticesTinyTerminalStaysInside`.

PR #184, opened 2026-08-02
fix(worktree): gate the worktree toggle when the target directory is not a repo

YoanWai:
## What

Spawning with worktree on resolved the target directory through `RepoRoot` and aborted the launch when that directory was not inside a git repository. A group whose path sits *above* a set of repos, an umbrella holding many checkouts, could not launch anything at all once `worktree_default` was on. The failure arrived after the prompt was already typed, reading `not inside a git repository: <path>` in the status bar.

The form and the quick bar now resolve the spawn directory the same way submit does, and gate the toggle on it:

- the value reads `unavailable (not a git repo)` in place of `on`/`off`
- pressing the toggle explains the refusal instead of flipping to a state that cannot work
- the session launches plainly in that directory

`spawnSession` keeps its hard error as the backstop.

Repo lookups are memoized per directory and dropped when the form or the quick bar opens, so the gate does not shell out to git on every frame while still seeing a repo created since the last look.

## Verification

- `gofmt -l .` prints nothing, `go vet ./...` clean, `go test ./...` and `go test -race ./internal/ui/` green on an isolated tmux socket.
- Drove the real umbrella path through the quick bar in a throwaway test: target dir `~/Desktop/projects/mreshet`, footer rendered `⇤ worktree: unavailable (not a git repo)`, submit spawned `cwd=~/Desktop/projects/mreshet worktreeRepo=""` with an empty error bar.

## Tests

`TestQuickWorktreeGatedInNonRepoGroup`, `TestFormWorktreeGatedInNonRepoDir` and `TestFormWorktreeStaysOnInRepoDir` cover the gate. Two existing quick-bar tests asserted the old behavior, that a worktree spawn into a non-repo dir surfaces an error and creates nothing; they now run against a real git repo, where the group default and the toggle override are what they actually mean.


YoanWai:
### Code review

Found 2 issues, both fixed in 18cbaa4:

1. The repo memo never expired, so a directory git-initialised while the quick bar stayed open kept reading as `unavailable` until the bar was closed and reopened. The bar is designed to stay active across prompts, so this made the toggle wrong in exactly the case where the user had just fixed the cause. Entries now carry the time they were taken and expire, which still keeps git off the repaint path.

https://github.com/YoanWai/agent-manager/blob/2b01985baf041d1a451822552f0d70d6008ea097/internal/ui/settings.go#L59-L73

2. `docs/usage.md` still promised the behavior this PR replaces: "A directory that is not a git repo, or a worktree that fails to create, blocks the spawn with an error instead of falling back to a shared directory." The gate makes that false for the non-repo half. The docs now describe the field reading `unavailable (not a git repo)`, the toggle explaining itself, and the session spawning plainly, with the umbrella-group case named.

https://github.com/YoanWai/agent-manager/blob/2b01985baf041d1a451822552f0d70d6008ea097/docs/usage.md#L58-L60

Checked and cleared: the `spawnSession` check still hard-errors for direct callers ([form.go](https://github.com/YoanWai/agent-manager/blob/18cbaa49a5a81009d5ed3d3a7c53c01eb4cc1ce0/internal/ui/form.go#L532-L552)), no comment in the touched files is contradicted or left stale, and nothing here repeats feedback from #175 or #177.

PR #185, opened 2026-08-02
fix(form): arrows move focus only, group picking moves to left/right

YoanWai:
Part 2 of #180, completing it (part 1 was #182).

Scrolling the new-session or new-group card with ↑/↓ used to fall into the group/parent picker: further presses silently changed the selection, and through the group's default path also overwrote the dir field mid-edit.

- ↑/↓ (and tab) now always move between fields, in both cards.
- The group/parent selection cycles with ←/→, matching the tool and worktree fields; the expanded picker stays visible under the field with its highlight following.
- The dir default still follows a deliberate group change while the field is untouched; a hand-edited dir is never overwritten.
- Hints updated: "←→ pick group · tab/↑↓ move".

Verified live in an isolated tmux run: ↓ cycles focus straight through the group field and wraps, selection only moves on ←/→. `go test ./...` green.

### DIFF
--- a/internal/ui/form.go
+++ b/internal/ui/form.go
@@ -309,11 +309,7 @@ func (m *Model) handleFormKey(msg tea.KeyMsg) (tea.Model, tea.Cmd) {
 		m.formFocus(-1)
 		return m, nil
 	case "up":
-		if m.form.focus == fieldGroup {
-			if !m.moveGroupCursor(-1) {
-				m.formFocus(-1)
-			}
-		} else if dirSuggesting {
+		if dirSuggesting {
 			if !m.pathSugg.move(-1) {
 				m.formFocus(-1)
 			}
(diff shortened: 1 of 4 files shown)


YoanWai:
### Code review

Two passes (key-routing/behavior bug scan, git-lineage + AGENTS.md + issue-coverage check). No defects in the diff; two follow-ups applied in-PR:

1. `moveGroupCursor`'s doc comment claimed a false result moves focus, which no caller does anymore; the function now wraps at the ends and the comment matches (e1a50a5, 380d89a).

2. Group/parent picking now wraps like the tool and worktree pickers, and the move-to-group modal reuses that wrap instead of reimplementing it.

https://github.com/YoanWai/agent-manager/blob/380d89a3cfe02a930629e54840c4be99be8712f4/internal/ui/form.go#L382-L390

Lineage checked: #99's escape-at-boundary fix is superseded, not regressed (arrows no longer enter the picker at all); #177's worktree-default follow still fires on deliberate picking. A hand-edited dir is never overwritten since only `dirAuto` state takes group defaults. Together with #182 this covers both complaints in #180.

PR #191, opened 2026-08-02
fix(groups): make group creation immediate and reliable

YoanWai:
## What changed

- update the in-memory group tree immediately after creation and select the new group
- reveal new groups through search, archived, empty-group, and collapsed-parent states
- store the path and worktree choice atomically and reject duplicate group creation
- exclude archived groups from destination pickers
- keep group-form fields sized to the modal
- add regression coverage for the full creation flow

## Why

A newly created group could remain invisible until the TUI restarted because the list depended on a later refresh and could still be filtered. Group creation also used an upsert followed by a separate worktree update, allowing duplicate submissions to overwrite existing settings and leaving persistence split across two writes.

## How it was verified

- [x] `go vet ./...` passes
- [x] `env -u TMUX TMUX_TMPDIR=/tmp/amtest go test -race ./...` passes
- [x] `go build ./...` passes

### DIFF
--- a/internal/store/store.go
+++ b/internal/store/store.go
@@ -15,7 +15,6 @@ import (
 // listed a moment earlier can tell that race apart from a real failure.
 var ErrSessionGone = errors.New("session no longer exists")
 
-// ErrGroupExists reports an add that would replace an existing group.
 var ErrGroupExists = errors.New("group already exists")
 
 type Session struct {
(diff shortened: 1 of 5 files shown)


YoanWai (commented):
Self-review complete. I checked the persistence boundary, duplicate behavior, local-to-polled state transition, filtered and collapsed list states, archived destinations, responsive sizing, and regression coverage. No blocking findings.

YoanWai:
Addressed the review nit in 52ff02e by removing the behavior-only comments. Focused race tests and vet remain clean.

PR #203, opened 2026-08-03
fix(status): prefer waiting over working matches

### LINKED ISSUE #112: Claude pane with an approval dialog reads as working, not waiting
Claude's status rules list its two `working` patterns above its `waiting` patterns, while codex and grok list `waiting` first. `Engine.Match` returns on the first matching rule (`internal/status/status.go:95-99`), so when a Claude pane shows an approval dialog while `esc to interrupt` is still on screen, the session reads as **working** instead of **waiting**.

Claude rules today (`internal/config/config.go:237-246`):

```toml
rules = [
  { state = "working", pattern = "(?m)^[✻✳✶✽✢·✦✧+*] \\S+… \\(" },
  { state = "working", pattern = "esc to interrupt" },
  { state = "waiting", pattern = "Enter to confirm" },
  { state = "waiting", pattern = "(?m)^[ \\x{A0}]*❯[ \\x{A0}]+\\d+\\." },
  { state = "errored", pattern = "(?im)^\\s*error:" },
]
```

codex (`config.go:284-294`) and grok (`config.go:310-319`) both put their dialog patterns first, which is the ordering that gives the right answer: a session blocked on the user should win over leftover activity text.

An agent parked on a permission prompt shows as busy, which is the exact state the manager exists to surface. Surfaced by a user in [this r/AIAgentsInAction thread](https://old.reddit.com/r/AIAgentsInAction/comments/1v9brwr/a_tmux_tui_for_running_coding_agents_live_status/p0o8ne6/).

Fix: move Claude's two `waiting` rules above its `working` rules so all four tools share the same precedence, and cover it with a pane fixture holding both a dialog and `esc to interrupt`.

Note `defaultConfig` changes only reach fresh installs: an existing user's on-disk config keeps the old ordering, so this needs a migration path rather than a defaults edit alone.

snowyukitty:
## What changed

- Prefer a matching `waiting` rule over an earlier `working` match within the current pane turn.
- Put Claude's built-in waiting rules before working rules for newly generated configs while preserving the fix for existing configs.
- Keep typed input outside the matching scope unless a real waiting footer is present.
- Add regression coverage for mixed approval panes, persisted rule order, modal footers, and numbered-input false positives.

## Why

An approval dialog can coexist with a stale spinner or `esc to interrupt` line. Returning the first working match makes the session look busy even though it is blocked on the user. The match-time precedence is required for existing `config.toml` files whose rule order is not migrated.

Closes #112

## How it was verified

- `go test ./internal/config ./internal/status`
- `go vet ./...`
- `go build ./...`
- `git diff --check`
- Every changed Go file passes `gofmt -l`.
- Synthetic merge trees with the current heads of #200 and #201 are conflict-free; those PRs touch adjacent config/test sections but do not implement this status precedence.
- `go test ./...` was also attempted on Windows. The touched packages pass, while unrelated packages hit existing Windows-specific assertions and Application Control blocks some temporary test executables.

- [ ] `gofmt -l .` prints nothing (the Windows CRLF checkout makes the repository-wide form non-actionable; all four changed files pass)
- [x] `go vet ./...` passes
- [ ] `go test -race ./...` passes (not run: this Windows host has no C compiler)

### DIFF
--- a/internal/config/config.go
+++ b/internal/config/config.go
@@ -202,7 +202,8 @@ func writeDefault(path string) error {
 const defaultConfig = `poll_interval = "2s"
 
 # Rules are matched top-down against the visible pane text (ANSI stripped);
-# first match wins. When no rule matches, the newest turn decides:
+# first match wins, except a matching waiting rule outranks a working match.
+# When no rule matches, the newest turn decides:
 # the content region is the text above the last activity_cutoff match
 # (the tool's input box). If the region's last content line — skipping
 # chrome_line matches (blanks, separators, input-box borders) — is a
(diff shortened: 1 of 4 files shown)


YoanWai (approved):
Thanks for this one! Ran the full race suite here plus a live A/B against real panes: the shipped binary reads a numbered draft during an active turn as waiting, this branch reads it correctly as working, and the mixed approval pane resolves to waiting as intended. Nice work on the footer scoping.

YoanWai:
Shipped in [v0.20.0](https://github.com/YoanWai/agent-manager/releases/tag/v0.20.0).

PR #205, opened 2026-08-03
feat(ui): filter sessions that need attention

YoanWai:
## Summary
- Press `w` to show only sessions that need attention (`waiting`, `finished`, `errored`); press again to show all.
- Header badge, session counts, and group rollups follow the filter; folds open so matches stay visible.
- Filter modes live in `statusFilterCycle` so more statuses can be added later without rewiring keys.

## Test plan
- [x] `env -u TMUX TMUX_TMPDIR=/tmp/amtest go test ./internal/ui/ -run StatusFilter`
- [ ] Manual: with mixed statuses, `w` shows only attention rows and `ATTENTION` badge; second `w` restores the full list
- [ ] Manual: folded group still reveals a waiting session under the filter
- [ ] Manual: `?` help and footer list the new binding

### DIFF
--- a/docs/usage.md
+++ b/docs/usage.md
@@ -33,7 +33,7 @@ Agent sessions live on a private tmux server named `agentmgr`, so they never mix
 | `s` | Settings (quick-spawn tool, theme, list density, review layout, after quick send, session keys, worktree sessions) |
 | `\|` | Resize the split: `←→` nudge the divider, `enter` commits, `esc` cancels |
 | `t` | Toggle archived view |
-| `f` | Filter to sessions that need attention (`waiting`, `finished`, `errored`); press again to show all |
+| `w` | Filter to sessions that need attention (`waiting`, `finished`, `errored`); press again to show all |
 | `e` | Hide / show empty groups |
 | `/` | Search |
 | `?` | Help |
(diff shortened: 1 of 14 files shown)


YoanWai:
@coderabbitai review

YoanWai:
@coderabbitai full review

YoanWai:
Review pass on this branch turned up five things, the first one matters most:

- **Bulk actions ignore the filter.** `sessionsInGroup` (group x/v), `reviveAllDead`, and kill-all still read `visibleSessions`, so with `f` active they act on sessions the list is hiding. 5 sessions, 4 working 1 waiting, filter to attention, press X: the confirm undercounts and kills the 4 hidden working sessions. Breaks sessionsInGroup's own invariant that a group action covers exactly the rows on screen.
- **A session spawned under the filter is invisible.** `submitForm` doesn't reset `statusFilter` (submitGroupForm does); a new session starts as starting/working, the attention filter rejects it, and the spawn looks like it failed.
- **Group rail count and roster disagree.** `viewGroupAgents` takes total from the filter-aware count but iterates unfiltered `visibleSessions`, so the rail lists agents the filter hides, and the truncation guard stops firing.
- **Filter empty-state overwrites the archive one.** With t + f active and an empty archive, the 't back to active' hint is gone and 'f show all' reveals nothing.
- **Attention set encoded twice.** `statusFilter.matches` re-lists waiting/finished/errored, which the poller also hardcodes; worth one shared definition so they can't drift.

PR #206, opened 2026-08-03
fix(ui): honor the default tool in new sessions

steveprentice:
## Summary

- initialize the New Session tool picker from the Settings default
- label the setting consistently as the default tool
- add a regression test for the selected tool

## Why

Quick spawn honored this setting, but the New Session form always selected the first enabled tool.

## Verification

- `gofmt -l .`
- `go vet ./...`
- `env -u TMUX TMUX_TMPDIR=/tmp/amtest-default go test -race ./...`
- `go build ./...`

### DIFF
--- a/internal/ui/form.go
+++ b/internal/ui/form.go
@@ -252,7 +252,7 @@ func (m *Model) enabledToolNames() []string {
 }
 
 func (m *Model) openForm() {
-	tools := m.enabledToolNames()
+	tools, toolIndex := m.defaultToolSelection()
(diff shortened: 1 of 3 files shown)


YoanWai:
@coderabbitai review

YoanWai (approved):
Thanks for the fix! Nice to have the form and quick spawn behaving the same.

YoanWai:
Shipped in [v0.20.0](https://github.com/YoanWai/agent-manager/releases/tag/v0.20.0).

PR #208, opened 2026-08-03
feat(ui): install updates from the settings version row

YoanWai:
## What changed

The settings version line is now a focusable field. When a newer release is available it becomes an action row (`↵ update to vX`), and Enter runs the same `applyUpdateCmd` path as `u` on the update notice in the messages modal (refresh latest tag, download, swap binary, restart). While the download runs the row shows progress and settings stay open; when already up to date, Enter still saves and closes.

## Why

Settings already showed `↑ vX available`, but it was display-only. Users who saw the badge there had to open messages and press `u` to install. One place should be enough.

## How it was verified

- [x] `go vet ./...` passes
- [x] `go test ./...` passes (isolated tmux socket)
- [x] Unit coverage for apply success, idle when current, and apply failure


YoanWai:
@coderabbitai review

YoanWai:
@coderabbitai full review

PR #212, opened 2026-08-03
feat(update): update in place from any install source

YoanWai:
## What

`u` on the update message (and `enter` on the Settings version row) now updates every install source instead of refusing package-manager installs with an error.

## How

- `internal/update/source.go`: `DetectManager` reports who owns the binary from the `main.buildSource` ldflag and the symlink-resolved install path: Homebrew Caskroom/Cellar/opt trees, mise's `installs/ubi-…-agent-manager` layout, and `pacman -Qqo` ownership on Linux. Everything else is a direct install.
- Package-manager installs suspend the TUI with `tea.ExecProcess` and hand the real terminal to the manager's own upgrade command (`brew upgrade --cask yoanwai/tap/agent-manager`, `brew upgrade agent-manager`, `mise upgrade --bump ubi:YoanWai/agent-manager`, `yay`/`paru -S agent-manager-bin`), so progress output and password prompts work as in a plain shell.
- Direct installs keep the verified download and atomic swap.
- On success the manager restarts into the new build through the PATH entry: brew and mise delete the version directory the running image was loaded from, so the resolved exec path of the old build can vanish mid-upgrade.
- A pacman install without an AUR helper surfaces the exact command to run instead of attempting anything.

## Verification

- `env -u TMUX TMUX_TMPDIR=/tmp/amtest go test ./...` green, `gofmt`/`go vet` clean.
- Live end-to-end on the Homebrew cask path: built this branch as 0.19.0 into the real Caskroom, pressed `u` on the update message, watched `brew upgrade --cask` run on the terminal, and the manager came back as 0.20.0 (`✦ Updated to 0.20.0` digest, process running `/opt/homebrew/bin/agent-manager`) with every session intact.
- mise layout verified against a real `mise use -g ubi:YoanWai/agent-manager` install (`installs/ubi-yoan-wai-agent-manager/<version>/agent-manager`).

### DIFF
--- a/docs/install.md
+++ b/docs/install.md
@@ -38,7 +38,7 @@ Run inside [WSL2](https://learn.microsoft.com/windows/wsl/install): agent-manage
 
 ## Updating
 
-The manager checks GitHub Releases once a day and shows a `↑ vX.Y.Z available` badge in the header when a newer version is out. Press `u` on the update message (or `enter` on the version row in Settings) and the manager updates itself whatever the install source: a Homebrew, mise, or AUR install hands the terminal to that package manager's upgrade command, a direct install downloads the release and swaps the binary in place, and either way the manager restarts into the new build with every session intact.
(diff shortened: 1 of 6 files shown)


YoanWai:
@coderabbitai review

YoanWai:
@coderabbitai review

PR #217, opened 2026-08-04
test(ui): pin alternate scroll off

YoanWai:
#215 turned DECSET 1007 off so wheel notches stop arriving as arrow keys that walk the session cursor, but nothing in CI held that in place. Three tests now do.

`TestDisableAlternateScrollEmitsReset` swaps `os.Stdout` for a pipe and asserts the exact bytes `\x1b[?1007l` go out. `TestStartupDisablesAlternateScroll` reads `ui.DisableAlternateScroll` out of `run()`'s syntax tree, since `run()` takes over the terminal and cannot be exercised headlessly. `TestNoAlternateScrollEnableInTree` walks the repo's Go sources for the enable sequence, so re-introducing it anywhere fails rather than silently restoring the old behavior.

All three were mutation-checked. Flipping the constant to `1007h` fails the first with `emitted "\x1b[?1007h", want "\x1b[?1007l"`. Removing the call from `run()` fails the second with `run() never calls ui.DisableAlternateScroll`. Adding the sequence to another file fails the third, naming that file. Full suite and `go vet` green.

### DIFF
--- a/internal/ui/altscroll_test.go
+++ b/internal/ui/altscroll_test.go
@@ -1,0 +1,71 @@
+package ui
+
+import (
+	"io"
+	"io/fs"
+	"os"
+	"path/filepath"
+	"strings"
+	"testing"
+)
+
+// The reset reaches the terminal verbatim: a wheel notch stays inert in the
+// list only while the terminal is not translating it into arrow keys.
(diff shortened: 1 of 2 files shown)


YoanWai:
@coderabbitai review

PR #222, opened 2026-08-04
docs(feed): announce v0.20 fork, filter, and Pi

YoanWai:
## Summary
- Fill the empty remote message feed with a v0.20 notice for installed clients
- Points at the release notes; credits @steveprentice and @snowyukitty

## Test plan
- [ ] `docs/messages.json` is valid JSON
- [ ] Banner/title/body lengths within feed limits
- [ ] After merge, a running agent-manager refresh shows the card (or after 6h cache)

### DIFF
--- a/docs/messages.json
+++ b/docs/messages.json
@@ -1,1 +1,16 @@
-[]
+[
+  {
+    "id": "v020-fork-filter-pi",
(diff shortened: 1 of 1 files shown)


YoanWai:
Superseded: v0.20 is two releases behind, and the feed slot goes to the next release instead.

Issue #227, opened 2026-08-06
feat: terminal sessions, a Terminals section, and MCP terminal tools


YoanWai:
## Why

Agent sessions are first class; a plain shell is not. Builds, `git`, log tails and one off commands belong next to the agent that needs them, in the same list, on the same tmux server. Agents want the same shell: today an agent runs its commands inside its own pane, where the output competes with the conversation for the screen.

### 1. `T` spawns a terminal session

`T` on any row opens a shell session in the group under the cursor. It is a session like any other: same store row, same status column, same `enter`, `A`, `x`, `v`, `a`, `u`, `d`, `r`, `m`, `K/J`.

- Backed by a `[tools.terminal]` block carrying an empty `command`. `tmux.Create` (`internal/tmux/tmux.go:73`) appends no window command when the command is empty, so the pane stays on `$SHELL`. `command = "fish"` opens a different shell.
- Going through a real tool block is what keeps revive, status classification and rename working, with no terminal branch of their own.
- `backfillToolDefaults` (`internal/config/config.go:114`) adds tool blocks missing from an existing `config.toml`, so installed users pick the block up on the next launch.
- `mcpreg.Style("terminal", "")` resolves to `none`, so the shell launches clean: no MCP flag, no prompt flag, no rename directive.
- The block stays out of the CLI pickers. `sortedToolNames` (`internal/ui/form.go:217`) drops it, which covers the `n` form, the quick prompt (`internal/ui/quick.go:125`) and the Settings default tool (`internal/ui/settings.go:294`). Rename keeps a terminal session on its own tool.
- Directory: the pane's live path through a new `tmux.PaneCurrentPath` (mirroring `PanePID`, `internal/tmux/tmux.go:372`), falling back to the stored `Cwd`. A group row uses the group default path.
- `default_status = "idle"`, and the name starts as a placeholder (`terminal-a1b2`) that `r` renames.
- The conversation keys (`space`, `f`, `ctrl+r`, `R`) say plainly that the row is a shell.

### 2. Terminals section, with an inline option

A Settings row `terminal placement`, cycled with `←→` alongside the existing rows (`internal/ui/modals.go:319`), stored under `terminal_placement` beside `worktree_default` and `list_density` (`internal/ui/keys.go:415`). A fresh install starts on `split`.

- `split`: one Terminals section under the whole tree, holding every shell from every group, each row naming the group it belongs to.

```
▸ backend
    claude-auth      working
    codex-migrate    idle
[...]

PR #232, opened 2026-08-07
feat(ui): spawn a plain shell session with T

andrelago13:
## What changed

`T` opens a shell tab in the group under the cursor. It is a session like any other — same list, same status column, same `enter`, `A`, `x`, `v`, `R`, `a`, `u`, `d`, `r`, `m`, `K/J` — with the user's shell in the pane instead of an agent.

- Backed by a `[tools.terminal]` block carrying an empty `command`, which `tmux.Create` already leaves on `$SHELL`. Going through a real tool block is what keeps revive, status classification and rename working with no terminal branch of their own, and `command = "fish"` opens a different shell.
- What marks it is `shell = true` on the block, never its name. `sortedToolNames` skips shells, so they stay out of the `n` form, the quick prompt and the Settings default tool; `T` resolves the shell through the flag. The flag is deliberately **not** backfilled by `mergeTool`, so a `[tools.terminal]` block someone already wrote stays the agent CLI they meant.
- `mcpreg.Style` resolves to `none` and no rename directive is armed, so the shell launches clean.
- Directory: the pane's live path through a new `tmux.PaneCurrentPath`, stat'd and falling back to the recorded `Cwd`. A group row uses the group default path.
- **The keys that write into a pane refuse a shell.** `space` (`submitQuick`) and the review screen's `C` (`sendAnnotations`) both deliver through `SendText`, which pastes *and* presses Enter — on a shell that turns a sentence meant for an agent into a command on the user's machine. Both now name the row as a shell and send nothing. `f` refuses in the same terms rather than naming a config field.

## Why

Builds, `git`, log tails and one-off commands belong next to the agent that needs them, in the same list and on the same tmux server, instead of in a terminal tab you have to go find.

Part of #227 — this is the `T` key and the mechanism under it. The Terminals section, the placement setting and the MCP terminal tools stay with @YoanWai to build on top, as discussed in the review.

The `o` editor key that this PR originally carried has moved out, as asked; it is written up in #233 and ready to open behind this one.

## How it was verified

```
gofmt -l .
go vet ./...
env -u TMUX TMUX_TMPDIR=/tmp/amtest go test -race ./...
```

Driven against real sessions: `T` on a group row and on a session row, kill and revive, rename, `R`, and the archived view.

The guard is pinned from both ends. `TestQuickPromptRefusesAShell` checks the refusal; `TestQuickPromptNeverRunsWhatIsTypedAtAShell` is @YoanWai's own repro — the quick prompt is asked to `touch` a marker file at a terminal row, and the test fails if the file ever appears. `TestSendAnnotationsRefusesAShell` and `TestForkRefusesAShellInItsOwnTerms` cover the other two keys, and `TestABlockNamedTerminalIsOnlyAShellWhenItSaysSo` pins the compatibility case.

- [x] `go vet ./...` passes

## Review notes still open

Two things I would rather you decided than have me change unilaterally:

- **`R` on a shell reads wrong.** It is safe (confirm-gated, relaunches the shell), but the confirm says "ends the running agent and leaves its conversation behind". `R` landed upstream after your review, so it arrived here with the merge. #227's acceptance wants it to say the row is a shell — happy to add that here or leave it to you.
- **The guard's truth lives in config, not on the session.** Rewrite `[tools.terminal]` into an agent block while shell sessions exist and those rows lose the refusal, silently. Narrow — `backfillToolDefaults` re-adds a deleted block, so it takes an active rewrite — but it is the flip side of the compatibility fix, and the failure mode is the bug you blocked on. The durable fix is recording shell-ness on the session row at spawn, which is also what #227 needs to identify terminal rows for the Terminals section and `list_terminals`. That is a store column on your schema, so I have not added it here.


docs/usage.md
```diff
@@ -55,6 +57,18 @@ Press `space` to dock a prompt bar at the bottom of the sidebar. The target foll
 
 ![answering a working Claude Code session from the prompt bar, without attaching](demo-space.gif)
 
+## Terminal tabs
+
+`T` opens a shell tab in the group under the cursor: a session like any other — same list, same status column, same `enter`, `x` and `v` — with your shell in the pane instead of an agent. It lands in the selected session's directory, or in the group's default path when a group row is selected, so the shell that runs the tests sits next to the agent that wrote them. Reviving one reopens the shell in that directory; the conversation keys (`space`, `f`, `ctrl+r`) have no agent to talk to and say so.
```
YoanWai: Small thing with a sharp edge: `space` does not say so, it runs.

I spawned a terminal on your branch and sent `touch /tmp/<marker>` from the quick prompt. The file appeared, the pane shows the command executed at a shell prompt, and the status bar stayed empty. `submitQuick` (`internal/ui/quick.go:244`) calls `SendText` without checking the target's tool, and `SendText` is a paste followed by `send-keys Enter` (`internal/tmux/tmux.go:206`).

`f` is genuinely fine, `validateForkSource` refuses an empty `fork_command`, so your instinct there was right. I have not checked `ctrl+r`.

Why I care: the quick prompt is where people type sentences at an agent, and on a terminal row a sentence becomes a command. Two ways out, either works for me. Refuse a shell target in `submitQuick`, or paste without the Enter (`Driver.Paste` exists for exactly that) and let the user press it themselves. `internal/ui/diffview.go:1109` wants the same guard.

internal/ui/form.go
```diff
@@ -214,8 +214,17 @@ func (m *Model) groupDefaultDir(group string) string {
...
 
+// sortedToolNames is every configured agent CLI in picker order. The
+// terminal tab is a shell rather than a CLI to spawn agents with, so it is
+// left out; its own key launches it, and a rename still keeps a terminal
+// session on it.
 func sortedToolNames(cfg config.Config) []string {
-	names := cfg.ToolNames()
+	names := make([]string, 0, len(cfg.Tools))
+	for _, name := range cfg.ToolNames() {
+		if name != terminalTool {
```
YoanWai: One compatibility snag: this keys off the literal name, so anyone who already has a `[tools.terminal]` block loses it from the `n` form, the quick prompt and the Settings default tool, with nothing said. `terminal` is a plausible name for a hand-rolled block, since the docs invite people to add their own tools.

Prefer a property over a reserved name: skip blocks whose `command` is empty, or add an explicit field on `config.Tool`. That also puts the intent in the config rather than at the call site.

internal/ui/terminal.go
```diff
@@ -0,0 +1,68 @@
...
+	tool := m.cfg.Tools[terminalTool]
+	sess := store.Session{
+		ID:     newID(),
+		Name:   terminalTool + "-" + newID()[:4],
+		Tool:   terminalTool,
+		Cwd:    dir,
+		Group:  m.contextGroup(),
+		Status: status.Starting,
+	}
+	if err := m.launchNewSession(sess, tool, tool.Command, launchOptions{}); err != nil {
```
YoanWai, line 61: Worth knowing rather than fixing: the env never reaches the pane. `tmux.Create` builds the env prefix only when the command is non-empty (`internal/tmux/tmux.go:85`), so `buildLaunch`'s `AGENT_MANAGER_SESSION_ID` is dropped for terminals. `agent-manager rename` from inside a terminal tab will not find its session.

I am happy with that for now. A shell that cannot act as the session is the safer default, and #227 wires the MCP side up properly. A line in the docs so nobody debugs it later would be enough.

internal/ui/terminal.go
```diff
@@ -0,0 +1,68 @@
...
+// to the directory it was created in; for a group, its default path.
+func (m *Model) rowDir() (string, bool) {
+	entry, ok := m.selectedRow()
+	if !ok {
+		return "", false
+	}
+	if entry.isGroup {
+		return resolveExistingDir(m.groupPaths[entry.group], m.groupDefaultDir(entry.group))
+	}
+	if path, err := m.tmux.PaneCurrentPath(entry.sess.ID); err == nil && path != "" {
```
YoanWai: Minor asymmetry: the group branch validates through `resolveExistingDir`, this one returns `ok` true on whatever tmux reports, with no stat. A directory deleted under a running shell still reports its old path, so the editor opens on nothing. Stat the pane path too, and fall back to `Cwd` when it is gone.

internal/ui/editor.go
```diff
@@ -0,0 +1,101 @@
...
+		}
+	}
+	return ""
+}
+
+// editorCommand appends the directory to an editor line. A line carrying
+// arguments ("code -n") goes through sh, so the directory stays one
+// argument however it is spelled.
+func editorCommand(line, dir string) *exec.Cmd {
+	if strings.ContainsAny(line, " \t") {
```
YoanWai: The directory is safe here, `"$@"` keeps it one argument whatever it contains. That was the right call.

The editor line is the part I would tighten. It lands in `sh -c` unsplit, and two of its four sources are environment variables, so an `.envrc` in a cloned repo can set `EDITOR` and then `o` runs it. Splitting the line into argv and calling `exec.Command(argv[0], argv[1:]...)` keeps the arguments-allowed behaviour without a shell.

This one travels to #233 with the rest of the editor key.

internal/ui/editor.go
```diff
@@ -0,0 +1,101 @@
...
+)
+
+// guiEditors are probed on PATH when nothing is configured, in the order
+// preferred. They open a project window of their own and return at once.
+var guiEditors = []string{"code", "cursor", "windsurf", "zed", "subl", "idea"}
+
+// terminalEditors draw in the terminal they are launched from. Started
+// detached they would have nowhere to paint, so they take the screen over
+// from the manager the way an attach does.
+var terminalEditors = map[string]bool{
```
YoanWai: An editor outside this map starts detached with no stdio, draws nowhere, and the status line still reports that it opened. `ed`, `joe`, or anyone's `edit` wrapper lands there.

You already keep a `guiEditors` list, so the test can invert: detach the ones known to open their own window, and hand the screen to everything else. A GUI editor run through `tea.ExecProcess` returns immediately and costs a repaint. A terminal editor started detached is simply broken.

Also for #233.

YoanWai (commented):
Thanks for this, and no need to apologise for the flow. The PR is a better starting point than the issue I had written.

The terminal half is the design I was after. Routing through a real tool block, so revive, status and rename need no terminal branch of their own, is cleaner than the `kind` field I had sketched in #227, and the tests cover the cases I would have asked for. I am taking the mechanism as it stands.

One thing to fix and one thing to move.

The fix: `space` on a terminal row executes what you type. I ran it on your branch to be sure, details inline on `docs/usage.md`. It is a small guard, and it is the only thing between this and a merge for me.

The move: the editor key deserves its own PR. I opened #233 for it and wrote the design up from yours, precedence and all, so it is credited and ready to go. Splitting it keeps this PR the foundation it already is, and #233 can land right behind it.

Heads up on where this is going. I rewrote #227 around your mechanism, and it now also wants a Terminals section, a Settings toggle between that and inline, and MCP tools so an agent can drive a shell. That part is mine to build on top of yours. If any of it looks fun, say so and it is yours.

CodeRabbit's three are worth doing, the `cmd.Wait` one especially. Everything below I checked against your branch rather than read off the diff, and nothing there is urgent except the first one.

YoanWai (changes requested):
Marking this changes requested so the PR state matches the review above. Nothing new here.

Two things to land it:

- guard `submitQuick` so `space` on a terminal row cannot execute what is typed (the `docs/usage.md` comment has the repro)
- move the editor key to its own PR against #233

The rest are notes, take or leave them. Ping me when the guard is in and I will re-review the same day.

andrelago13:
@YoanWai thanks for the quick review, I've iterated on this with my agent following your guidance, let me know what you think. As you suggested I've separated out the other change, happy to follow-up on it once this PR is in!

YoanWai (dismissed):
Dismissed - superseded by a later review.

YoanWai:
Merged, thank you.

The `shell = true` flag was the better answer. I suggested keying off an empty command; a block being a shell only where its author said so is what actually holds, and keeping it out of `mergeTool` is the detail that makes it safe.

One follow-up on top in #241: `T` spawns on the keystroke, so a held key autorepeated into a pile of shells. There's a repeat window now, and I trimmed the footer on shell rows.

#233 is yours whenever you want it.

Thanks again for taking the time on this.

andrelago13:
Thank you, I'll tackle that early next week!

PR #236, opened 2026-08-07
chore: add CodeRabbit review config

YoanWai:
Adds `.coderabbit.yaml` so CodeRabbit reviews follow the conventions in AGENTS.md rather than its dashboard defaults.

`path_instructions` cover:
- **`**/*.go`** — comments explain a non-obvious why; speculative branches for shapes that cannot occur are defects; errors propagate rather than fall back silently; names spell out intent.
- **`**/*_test.go`** — tmux-touching tests run on an isolated socket via `env -u TMUX TMUX_TMPDIR=/tmp/<short>`; a missing `env -u TMUX` reaches the live server.
- **`internal/tmux/**`** — flag anything resolving to the default socket.
- **`internal/ui/**`** — `Update` stays non-blocking; I/O belongs in a `tea.Cmd`.
- **`internal/config/**`** — a behavior change that must reach existing installs needs a new field, since a stored config.toml keeps its old defaults.
- **`.github/workflows/**`** — actions pinned to a commit SHA.

Also: drafts skipped, poem off, `dist/`, `docs/demo/` and `go.sum` filtered from review.

Validated against CodeRabbit's published `schema.v2.json`.

## Summary by CodeRabbit

* **Chores**
  * Added automated review configuration for the main development branch.
  * Enabled review summaries, status reporting, and automatic responses.
  * Added tailored review guidance for Go, tests, terminal workflows, UI, configuration, and GitHub workflows.
  * Configured selected path exclusions and local knowledge-base learning.

### DIFF
--- a/.coderabbit.yaml
+++ b/.coderabbit.yaml
@@ -11,11 +11,9 @@ reviews:
   auto_review:
     enabled: true
     drafts: false
-    base_branches:
-      - main
   path_filters:
-    - "!dist/**"
-    - "!docs/demo/**"
+    - "!**/*.gif"
+    - "!**/*.png"
     - "!go.sum"
   path_instructions:
     - path: "**/*.go"
@@ -33,10 +31,11 @@ reviews:
     - path: "**/*_test.go"
       instructions: |
         Tests live beside the file they cover (listview.go -> listview_test.go).
-        Any test that starts a tmux server must run on an isolated socket via
-        env -u TMUX TMUX_TMPDIR=/tmp/<short>; $TMUX overrides TMUX_TMPDIR, so a
-        test that omits env -u TMUX reaches the user's live tmux server. Flag
-        kill-server or kill-session against a non-isolated socket as critical.
+        Every tmux invocation must name its socket explicitly with -L or -S; an
+        explicit socket is sufficient on its own. A bare tmux call resolves
+        through $TMUX to the developer's live server, so flag one that carries
+        neither an explicit socket nor env -u TMUX as critical, and likewise a
+        kill-server or kill-session that is not pinned to an isolated socket.
     - path: "internal/tmux/**"
       instructions: |
         This package owns the dedicated agentmgr socket. Flag anything that can
--- a/.coderabbit.yaml
+++ b/.coderabbit.yaml
@@ -1,0 +1,65 @@
+language: en-US
+early_access: false
+
+reviews:
+  profile: assertive
+  request_changes_workflow: false
+  high_level_summary: true
+  poem: false
+  review_status: true
+  collapse_walkthrough: true
+  auto_review:
+    enabled: true
+    drafts: false
+    base_branches:
+      - main
+  path_filters:
+    - "!dist/**"
+    - "!docs/demo/**"
+    - "!go.sum"
+  path_instructions:
+    - path: "**/*.go"
+      instructions: |
+        Comments are rare and explain a non-obvious why, never what the code
+        does. Do not ask for doc comments, JSDoc-style blocks, or "what" comments
+        on exported identifiers.
+        Reject speculative branches: a fallback or nil-guard for a shape that
+        cannot occur is a defect, not safety. Guards on genuinely external or
+        untrusted input are correct.
+        Errors propagate or are handled explicitly; silent fallbacks that mask a
+        failure are a defect.
+        Names spell out intent: no one-letter variables outside tight loop
+        indices, no invented abbreviations.
+    - path: "**/*_test.go"
+      instructions: |
+        Tests live beside the file they cover (listview.go -> listview_test.go).
+        Any test that starts a tmux server must run on an isolated socket via
+        env -u TMUX TMUX_TMPDIR=/tmp/<short>; $TMUX overrides TMUX_TMPDIR, so a
+        test that omits env -u TMUX reaches the user's live tmux server. Flag
+        kill-server or kill-session against a non-isolated socket as critical.
+    - path: "internal/tmux/**"
+      instructions: |
+        This package owns the dedicated agentmgr socket. Flag anything that can
+        resolve to the default tmux socket, and any global option or key binding
+        written to a server this process does not own.
+    - path: "internal/ui/**"
+      instructions: |
+        Bubble Tea: one Model, files grouped by feature. Update must not block;
+        flag synchronous I/O, exec, or sleeps on the update path instead of a
+        tea.Cmd.
+    - path: "internal/config/**"
+      instructions: |
+        Existing installs keep their stored config.toml, so editing a
+        defaultConfig value never reaches current users. A behavior change that
+        must reach them needs a new field, not an edited default.
+    - path: ".github/workflows/**"
+      instructions: |
+        Pin third-party actions to a commit SHA. Flag any workflow that exposes
+        secrets to pull_request_target or to a step running untrusted PR code.
+
+chat:
+  auto_reply: true
+
+knowledge_base:
+  learnings:
+    scope: local
--- a/.coderabbit.yaml
+++ b/.coderabbit.yaml
@@ -3,9 +3,9 @@ early_access: false
 
 reviews:
   profile: assertive
-  request_changes_workflow: false
+  request_changes_workflow: true
   high_level_summary: true
-  poem: false
+  poem: true
   review_status: true
   collapse_walkthrough: true
   auto_review:


.coderabbit.yaml
```diff
@@ -0,0 +1,65 @@
...
+        failure are a defect.
+        Names spell out intent: no one-letter variables outside tight loop
+        indices, no invented abbreviations.
+    - path: "**/*_test.go"
+      instructions: |
+        Tests live beside the file they cover (listview.go -> listview_test.go).
+        Any test that starts a tmux server must run on an isolated socket via
+        env -u TMUX TMUX_TMPDIR=/tmp/<short>; $TMUX overrides TMUX_TMPDIR, so a
+        test that omits env -u TMUX reaches the user's live tmux server. Flag
+        kill-server or kill-session against a non-isolated socket as critical.
```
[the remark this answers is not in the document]
YoanWai: Verified and applied. `internal/tmux/tmux.go:62` and `tmux_test.go:27` both prepend `-L <socket>`. A probe with $TMUX pointing at the live agentmgr server plus `TMUX_TMPDIR=/tmp/crprobe` resolved `tmux -L crprobe` to `/private/tmp/crprobe/tmux-502/crprobe`, so the explicit socket wins and $TMUX does not. The rule now requires an explicit `-L`/`-S` or `env -u TMUX`, and flags only invocations carrying neither.

.coderabbit.yaml
```diff
@@ -0,0 +1,64 @@
...
+        explicit socket is sufficient on its own. A bare tmux call resolves
+        through $TMUX to the developer's live server, so flag one that carries
+        neither an explicit socket nor env -u TMUX as critical, and likewise a
+        kill-server or kill-session that is not pinned to an isolated socket.
+    - path: "internal/tmux/**"
+      instructions: |
+        This package owns the dedicated agentmgr socket. Flag anything that can
+        resolve to the default tmux socket, and any global option or key binding
+        written to a server this process does not own.
+    - path: "internal/ui/**"
```
YoanWai, line 44: **The tmux socket rule guards the one package that cannot violate it.**

Every call in `internal/tmux` goes through `Driver.args`, which prepends `-L` (`tmux.go:62`), so the rule scoped here and to `**/*_test.go` has nothing to catch. Meanwhile the only bare tmux invocation in the product sits at `internal/ui/theme.go:425`:

```go
_ = exec.Command("tmux", "set-option", "-p", "allow-passthrough", "on").Run()
```

and `internal/ui/**` is reviewed only against "Update must not block". Widen that call to `set-option -g`, or add a `kill-session`, and the shipped binary reaches the user's own default-socket server with a green review. The rule belongs in `**/*.go`.

YoanWai (commented):
Self-review of this config, since CodeRabbit hit its rate limit here and the last commit merged unreviewed. Ten findings survived an independent verify pass. All of them are fixed in #238.

.coderabbit.yaml
```diff
@@ -0,0 +1,64 @@
...
+    enabled: true
+    drafts: false
+  path_filters:
+    - "!**/*.gif"
+    - "!**/*.png"
+    - "!go.sum"
+  path_instructions:
+    - path: "**/*.go"
+      instructions: |
+        Comments are rare and explain a non-obvious why, never what the code
```
YoanWai, line 21: **Two unset defaults override this instruction.**

`reviews.finishing_touches.docstrings.enabled` defaults to `true` and `reviews.pre_merge_checks.docstrings.mode` defaults to `warning`. A path instruction cannot suppress a separate top-level key, so every Go PR still draws a Docstring Coverage warning and an offer to write doc-comment blocks onto exported identifiers — exactly what this rule and AGENTS.md forbid. Both need to be set off explicitly.

.coderabbit.yaml
```diff
@@ -0,0 +1,64 @@
...
+  poem: true
+  review_status: true
+  collapse_walkthrough: true
+  auto_review:
+    enabled: true
+    drafts: false
+  path_filters:
+    - "!**/*.gif"
+    - "!**/*.png"
+    - "!go.sum"
```
YoanWai, line 17: **`!go.sum` silently disables golangci-lint.**

The schema says path_filters "also apply to 'git sparse-checkout' ... ignoring excluded ones (starting with '!') when cloning the repository", and `reviews.tools['golangci-lint'].enabled` defaults to `true`. Without go.sum the module does not build — extracting the tree and removing it gives `missing go.sum entry for module providing package github.com/BurntSushi/toml` across eight packages — so the linter has nothing to run against and a PR carrying an errcheck or nilness bug merges with a clean report.

.coderabbit.yaml
```diff
@@ -0,0 +1,64 @@
...
+      instructions: |
+        This package owns the dedicated agentmgr socket. Flag anything that can
+        resolve to the default tmux socket, and any global option or key binding
+        written to a server this process does not own.
+    - path: "internal/ui/**"
+      instructions: |
+        Bubble Tea: one Model, files grouped by feature. Update must not block;
+        flag synchronous I/O, exec, or sleeps on the update path instead of a
+        tea.Cmd.
+    - path: "internal/config/**"
```
YoanWai, line 49: **`internal/store` has the same hazard and no rule.**

This rule exists because a stored config keeps its old defaults. `internal/store/store.go:113-139` is an append-only migrations slice whose loop swallows `duplicate column` (`:140-146`). Edit an existing entry — say `ALTER TABLE sessions ADD COLUMN acked ... DEFAULT 0` to `DEFAULT 1` — and the ALTER re-runs on upgraded databases, fails, gets swallowed, and fresh installs diverge from upgraded ones with nothing surfaced.

.coderabbit.yaml
```diff
@@ -0,0 +1,64 @@
...
+      instructions: |
+        Existing installs keep their stored config.toml, so editing a
+        defaultConfig value never reaches current users. A behavior change that
+        must reach them needs a new field, not an edited default.
+    - path: ".github/workflows/**"
```
YoanWai, line 54: **The self-update path gets no rule while the only security rule is workflow-scoped.**

`internal/update/apply.go` fetches a release asset, verifies it against a sha256 from `checksums.txt`, and renames the result over the live executable. A PR that drops `verifyChecksum`, follows a redirect to another host, or relaxes the staged file's permissions is reviewed only under the generic Go rules about comments and naming — and it ships to every user's binary.

.coderabbit.yaml
```diff
@@ -0,0 +1,64 @@
...
+      instructions: |
```
YoanWai, line 55: **This fires against our own convention.**

All five `uses:` in the repo are tag-pinned (`actions/checkout@v7`, `actions/setup-go@v7`, `codecov/codecov-action@v7`), and `.github/dependabot.yml` has a `github-actions` ecosystem bumping them weekly. Demanding 40-char SHAs means a comment on every workflow PR asking for a change we never intend to make. Flagging mutable refs (`@main`, no ref) is the rule that actually protects us.

.coderabbit.yaml
```diff
@@ -0,0 +1,64 @@
...
+        Comments are rare and explain a non-obvious why, never what the code
+        does. Do not ask for doc comments, JSDoc-style blocks, or "what" comments
+        on exported identifiers.
+        Reject speculative branches: a fallback or nil-guard for a shape that
+        cannot occur is a defect, not safety. Guards on genuinely external or
+        untrusted input are correct.
+        Errors propagate or are handled explicitly; silent fallbacks that mask a
+        failure are a defect.
+        Names spell out intent: no one-letter variables outside tight loop
```
YoanWai, line 29: **The one-letter ban contradicts the codebase.**

Non-test code has 391 `m`, 53 `d`, 52 `s`, 19 `p`, 15 `c` and 10 `w` method receivers, and AGENTS.md's Style section carries no naming rule at all. At `profile: assertive` this asks authors to rename `func (m *Model)` to `func (model *Model)` and diverge from every neighbouring method. Receivers and the usual short Go names need an explicit exemption.

.coderabbit.yaml
```diff
@@ -0,0 +1,64 @@
...
+early_access: false
+
+reviews:
+  profile: assertive
+  request_changes_workflow: true
+  high_level_summary: true
+  poem: true
+  review_status: true
+  collapse_walkthrough: true
+  auto_review:
```
YoanWai, line 11: **Review stops after five commits while the check stays green.**

`enabled` and `drafts` are both already the schema defaults, so this block encodes nothing — but the key it omits, `auto_pause_after_reviewed_commits`, defaults to `5`. On a PR that takes more than five cycles (routine for the UI work here) commit six onward is never reviewed, and the earlier passing status stays visible, so unreviewed code merges under a check that reads as approved.

.coderabbit.yaml
```diff
@@ -0,0 +1,64 @@
...
+        does. Do not ask for doc comments, JSDoc-style blocks, or "what" comments
+        on exported identifiers.
+        Reject speculative branches: a fallback or nil-guard for a shape that
+        cannot occur is a defect, not safety. Guards on genuinely external or
+        untrusted input are correct.
+        Errors propagate or are handled explicitly; silent fallbacks that mask a
+        failure are a defect.
+        Names spell out intent: no one-letter variables outside tight loop
+        indices, no invented abbreviations.
+    - path: "**/*_test.go"
```
YoanWai, line 31: **Generated tests are not bound by this rule.**

`reviews.path_instructions` govern review, not generation. `reviews.finishing_touches.unit_tests.enabled` defaults to `true` while `code_generation.unit_tests.path_instructions` defaults to `[]`, so accepting the generate-tests offer can produce a test that shells out to bare `tmux`. The next `go test ./...` then resolves through `$TMUX` to the live server and tears down the developer's own sessions.

.coderabbit.yaml
```diff
@@ -0,0 +1,64 @@
...
+      instructions: |
+        Tests live beside the file they cover (listview.go -> listview_test.go).
+        Every tmux invocation must name its socket explicitly with -L or -S; an
```
YoanWai, line 34: **This relaxation is overruled by AGENTS.md.**

`knowledge_base.code_guidelines.enabled` defaults to `true` and its default file patterns include `**/AGENTS.md`, which states `env -u TMUX` is mandatory. The reviewer therefore holds both rules at once, so a PR touching `internal/tmux/tmux_test.go:27` or `internal/ui/main_test.go:25` — helpers that already pass `-L testSocket` — can still be flagged critical for the missing `env -u TMUX`. The two sources have to say the same thing: the suite command carries `env -u TMUX`, and a helper with an explicit socket is already pinned.

PR #242, opened 2026-08-07
fix(ui): keep focus-mode scroll from losing rows under fast wheel input

redeye1011:
## Symptom

In focus mode (Enter on a session), scrolling up with the mouse wheel made
rows vanish from the bottom upward until the pane went blank. It was worst
during fast wheel bursts and when scrolling deep into history (offsets
~800+).

Reproduced and verified in: Ghostty terminal, both on the local machine and
over SSH, with agent-manager 0.23.0. Sessions tested: agy, codex, and
claude (with `/tui default`). After this patch, focused scrolling up no
longer loses lines in any of those tools, with and without SSH.

## Root cause

Focused scrolling issues `capture-pane` per wheel notch through the tmux
control client, with two defects:

1. The capture window is sized from `m.pane.box.height`, which can be stale
   relative to the real preview viewport (`previewPaneHeight()`), so the
   request window doesn't match the pane being painted.
2. Replies were applied whenever the offset matched, even if the pane was
   resized while the capture was in flight, and failed captures silently
   dropped the frame.

Under fast wheel input, replies arrive behind the watcher's live captures;
stale or wrongly-sized frames get painted and the viewport loses rows from
the bottom up.

## Fix

- `focusRegionCmd` sizes the capture window from `previewPaneHeight()`.
- `focusScrollMsg` now carries the requested row count and an `ok` flag; a
  reply captured before a resize, or one whose offset no longer matches, is
  rejected and the final viewport is recaptured instead of painting a stale
  frame.
- Failed captures no longer silently clear the preview.
- Typing-to-resume uses the same request path as scrolling.

## Verification

- Live TUI: the bug no longer reproduces in Ghostty, local and over SSH,
  across agy / codex / claude sessions.
- `gofmt -l .`, `go vet ./...`, `go build ./...`,
  `go test -race ./...` all pass (tmux 3.7b, Go 1.26.5).
- New regression tests: `TestFocusScrollRecapturesAfterPreviewResize`,
  `TestFocusScrollKeepsDeepHistoryFrame`.

## Notes for the reviewer

An earlier experiment also coalesced wheel input to one in-flight capture.
It was dropped: it freezes `TestScrollStopsAtHistoryTop` (the wheel stalls
whenever a capture reply isn't routed), and the offset-guard recapture in
this patch already prevents stale frames without suppressing input.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved focus recovery while viewing scrollback history.
  * Focused content now stays aligned after resizing the preview pane.
  * Prevented stale scroll captures from displaying outdated content.
  * Improved navigation through deep scrollback while preserving the expected pane height.
  * Added safeguards to ensure captured results apply only to the currently selected session.

### DIFF
--- a/internal/ui/focuskeys.go
+++ b/internal/ui/focuskeys.go
@@ -172,7 +172,7 @@ func (m *Model) handleFocusKey(msg tea.KeyMsg) (tea.Model, tea.Cmd) {
 	var resume tea.Cmd
 	if m.scrolledBack() {
 		m.focusScroll = 0
-		resume = m.focusRegionCmd(sess.ID, 0)
+		resume = m.requestFocusRegion(sess.ID)
 	}
 	if msg.Paste {
 		if err := pasteFocused(m.tmux, sess.ID, string(msg.Runes)); err != nil {
--- a/internal/ui/focusscroll.go
+++ b/internal/ui/focusscroll.go
@@ -17,7 +17,9 @@ const focusScrollStep = 3
 type focusScrollMsg struct {
 	sessID  string
 	offset  int
+	rows    int
 	preview string
+	ok      bool
 }
 
 // Mouse button codes as the reports carry them: the two wheel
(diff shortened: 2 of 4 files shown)


YoanWai:
Belated thanks, this one slipped through without a note. The diagnosis was right: `pane.box.height` going stale versus the real preview viewport, plus painting in-flight captures after a resize, is exactly how fast wheel bursts blanked the pane from the bottom up. Sizing from `previewPaneHeight()` and rejecting mismatched replies (then recapturing) is the right shape, and dropping the coalesce experiment was correct too. Shipped in [v0.25.0](https://github.com/YoanWai/agent-manager/releases/tag/v0.25.0).

PR #246, opened 2026-08-07
feat(ui): add OS dark/light mode auto-detection for themes

mikaoelitiana:
Adds automatic theme switching based on the operating system's light/dark mode preference.

## Features
- Detect OS-level dark/light mode on macOS, Windows, and Linux
- New 'theme auto-detect' setting in Settings (s key)
- When enabled, automatically selects 'classic' (dark) or 'solarized light' theme
- When disabled, uses manually selected theme (default behavior)
- Toggle with left/right keys in Settings

## Platform Support
- **macOS**: Dark
- **Windows**: Registry query for 
- **Linux**: XDG Desktop Portal, gsettings, GTK theme detection, environment variables

## Implementation
- New package:  - Platform-specific detection logic
- New test file:  - Comprehensive tests
- Modified: , , , 

## Notes
- Opt-in feature (disabled by default)
- Non-invasive: when disabled, behavior is exactly as before
- Only affects agent-manager's own UI, not the agent CLIs
- Theme detection runs once at startup

Fixes #245

## Summary by CodeRabbit

- **New Features**
  - Added automatic theme selection based on the operating system’s light or dark mode preference.
  - Added a settings option to enable or disable automatic theme detection.
  - Automatic theme detection supports macOS, Windows, and Linux environments.
  - Disabling automatic selection restores the manually selected theme.
  - The automatic theme preference is saved between sessions.

- **Bug Fixes**
  - Added fallback behavior when system theme detection is unavailable, unsupported, or fails.

PR #251, opened 2026-08-08
feat(ui): follow the OS light/dark preference with theme auto-detect

### LINKED ISSUE #245: feat: auto-switch theme based on OS light/dark mode preference
## Feature Request

Add automatic theme switching for agent-manager's UI based on the operating system's light/dark mode preference.

### Current Behavior
- Theme must be manually selected via Settings (`s` key)
- 15 built-in themes (9 dark, 6 light) are available
- No automatic detection of system color scheme

### Proposed Behavior
- Detect OS-level dark/light mode preference at startup
- Automatically select an appropriate theme (e.g., switch between 'classic' dark and 'solarized light' or 'paper')
- Allow manual override in Settings
- Optionally: watch for system theme changes while running and switch dynamically

### Notes
- This would only affect agent-manager's own UI (sidebar, header, diff view, etc.)
- Individual agent CLIs (Claude Code, OpenCode, etc.) run in their own tmux panes and maintain their own theming independently
- agent-manager already syncs terminal background to match its theme via OSC 11, so this would extend that automation to the theme selection itself

### Platforms
Should work on:
- macOS (via `defaults read -g AppleInterfaceStyle` or `NSRequiresAquaSystemAppearance`)
- Windows (via registry query)
- Linux (via environment variables or dbus)

YoanWai:
Adds a **theme follows OS** setting (off by default) that resolves the palette at startup from the environment's color scheme. Builds on the work in #246 by @mikaoelitiana (credited as co-author) and on the dark capture backdrop from #250.

## Detection chain
1. **macOS**: `defaults read -g AppleInterfaceStyle` — the key only exists in dark mode, so a nonzero exit is the light answer and only a missing binary reads unknown.
2. **Linux desktops**: XDG desktop portal (`org.freedesktop.appearance color-scheme`, speaks for GNOME/KDE/etc.), then `gsettings` color-scheme, then a dark-named GTK theme.
3. **Any terminal** (covers SSH and bare servers): OSC 11 background query on the controlling tty — answered by the terminal directly, or by tmux on the pane's behalf from its client's terminal — then the COLORFGBG convention. Luminance picks the side.
4. Unknown changes nothing: the stored theme stays.

Every command runs under a 500ms cap and the tty query under a 200ms select deadline, so a broken desktop bus or a silent terminal costs a beat of startup, never a hang.

## Behavior
- Polarity-preserving: a stored theme already on the detected side stays; only a mismatch switches to `classic` / `solarized light`.
- The `theme` key always keeps the manual choice; `theme_auto` is its own key. Turning auto off returns to the manual theme, live.
- Stepping the theme picker by hand turns auto off — a manual choice wins over silent override on next start.
- Captured agent panes keep the pinned dark backdrop from #250 under any light theme, so auto-selecting light never washes out the embedded CLIs.

## Verification
- Unit tables for every platform branch (mock runner), the OSC 11 parser (16/8/4-bit channels, BEL/ST), COLORFGBG, and polarity resolution; settings persistence covered end to end.
- Live on this machine (macOS dark): `Detect()` returns dark; a seeded store with `theme='solarized light', theme_auto=on` boots into classic, settings show `theme follows OS ◂ on ▸`, toggling off restores solarized light immediately and persists `theme_auto=off` with the manual theme intact.
- Live OSC 11 round trip against tmux 3.7b on an isolated socket: dark pane style answers `rgb:0f0f/1111/1515` → parsed 15/17/21; with no client and default colors the query stays silent and the 200ms deadline returns unknown, no hang.
- `go build` for darwin/arm64 and linux/amd64 (the shipped targets), `go vet`, full test suite green.

Fixes #245


YoanWai:
@coderabbitai review

PR #260, opened 2026-08-10
feat(ui): step in and out of a session with the arrow keys

YoanWai:
Right on a list row steps in: a session is focused, a collapsed group opens. Left steps out: an open group closes, and inside a focused pane - when the agent's caret sits at the head of its prompt - it returns to the list.

## How Left stays safe inside the prompt

Left only leaves focus where the agent would do nothing with it: the caret at the head of its prompt, with nothing but the prompt marker to its left. Anywhere else it moves the caret as usual, so no keystroke is lost.

The check reads the tool's existing `activity_cutoff` marker against the pane row under the caret (the focus watcher already streams the caret cell), and every cell between marker and caret must be blank. Claude pads its marker with a non-breaking space, so blank means any space rune, not the ASCII one alone. Wrapped prompt rows carry no marker, so Left there reaches the agent and lands at the end of the row above.

Tools without a row-anchored marker (terminals, pi) never match, and Left forwards exactly as before. Unknown tools take the same safe path.

## Verified

- Live against a real Claude pane: at an empty prompt the caret sits at column 2 and Left is a no-op for Claude; with text typed, Left walks the caret back cell by cell and only the press at the head returns to the list, with the draft intact.
- `go vet ./...`, `gofmt -l`, and the full suite green; new tests cover the caret walk (nbsp padding, wrapped rows, quoted markers, wide markers, indented composers, stale and scrolled panes), Left/Right routing in focus and list modes, and the `InputPrefix` row anchoring.

## Summary by CodeRabbit

- **New Features**
  - Added optional ←→ keyboard navigation for stepping into sessions, opening or closing groups, and returning from focused prompts.
  - Added a persisted Arrow Step setting, enabled by default and labeled as beta.
  - Added an in-app beta notice with feedback access.

- **Documentation**
  - Updated the usage guide and help screen with the new keyboard shortcuts.

- **Bug Fixes**
  - Left-arrow navigation now exits focus only when the cursor is at the prompt start, preserving agent shortcuts such as Alt+Left.


YoanWai:
Added since the first review: the pair ships on but marked beta — Settings gains a yellow-tagged "←→ step in/out" row (persisted as `arrow_step_keys`), off disables right-to-focus, left-to-fold and left-to-unfocus together, and a built-in message introduces the keys, notes they are under test, and links a prefilled report scoped to the beta area. Zero-width `activity_cutoff` matches are rejected in `InputPrefix` and the behavior-only comment is gone.

YoanWai:
@coderabbitai The zero-width `activity_cutoff` guard has been in since 3b399e8 (`loc[1] == 0` in InputPrefix, regression case "zero-width cutoff" in inputprefix_test.go); the re-surfaced comment is stale. The Left-key encoding case and the test comment are addressed in the latest commit.

PR #269, opened 2026-08-11
feat(contrib): add cmux and desktop notification pipeline for session state changes

redeye1011:
## What changed

Adds `contrib/notifications/` — a self-contained notification pipeline that turns agent-manager's per-session status files (`waiting` / `working` / `finished` / `idle`) into correctly-routed notifications for users running agents under [cmux](https://github.com/manaflow-ai/cmux) (Ghostty-based macOS terminal), locally or over ssh, and for users at a Linux desktop.

Two components per platform, one integration point (the status files agent-manager already writes — no agent-specific configuration, covers every managed agent type):

- **Watcher** (`am-status-notify`): polls the status directory every 2s, fires on flips to `waiting`/`finished`, seeds silently on start, dedupes per session. Notification title is the session's name in agent-manager (read from `state.db` via python3 stdlib `sqlite3`, falling back to the raw id), body is `Waiting for your input` / `Finished` — mirroring cmux's native Claude Code notification shape. Lifecycle via systemd user unit (Linux) or launchd agent (macOS); `install.sh`/`uninstall.sh` for both.
- **Delivery module** (`cmux-notify`):
  - **macOS**: writes an OSC 777 escape to the agent-manager process's TTY (probed live via `ps`). cmux/libghostty raises it natively — pane ring, sidebar badge, notification panel, macOS banner — attributed to the correct workspace. Single channel; deliberately no `osascript` side channel (those banners are attributed to "Script Editor", can't focus the pane, and duplicate cmux's own).
  - **Linux**: `notify-send` to the freedesktop daemon (terminal-agnostic — works regardless of which terminal emulator agent-manager runs in), plus the cmux relay shim with explicit `--workspace` targeting for ssh workspaces, the workspace id read live from the agent-manager process environment per event.
  - Events spool to a TSV FIFO (cap 200) only when all channels fail, and replay in order; a stored target that died while queued gets one retry against the live-probed target.
- **Optional Codex per-turn pings** on Linux via codex's native `notify` hook pointing at the delivery module; payload JSON is normalized to `last-assistant-message`, one line, ≤160 chars.
- **Tests**: 18-check sandboxed interface suite for the Linux delivery module (targeting precedence, spool format, FIFO replay, stale-target retry, cap, body normalization).

[...]

PR #276, opened 2026-08-12
feat(ui): desktop notifications when a session needs you

### LINKED ISSUE #226: feat: desktop notifications when a session needs you
## Why

herdr users repeatedly cite **notifications** (terminal bell / OS ping when an agent blocks) as a primary reason to switch from raw tmux. We already classify `waiting` / `errored` / `finished` and filter with `w`. Status without a ping still forces the user to watch the list.

Related but different: #77 (in-manager informative toasts with actions, all-clients-or-nothing). This issue is the **thin ASAP slice**: OS / terminal notifications when attention is needed, without solving the full toast system.

## Goal

When a session transitions into a state that needs the human (at least `waiting` and `errored`), fire a desktop/terminal notification so the user can look away from the manager.

## Requirements

- Fire on transition into attention-worthy status, not on every poll tick (no spam while status stays `waiting`).
- At minimum: `waiting`, `errored`. Optionally `finished` behind a setting (default off or on TBD).
- Respect focus: if the manager is focused and that session is selected, prefer soft/no OS noise (investigate).
- Works on macOS and Linux; Windows via WSL should degrade gracefully (document if OS notify is weak).
- One settings toggle (and maybe per-status checkboxes later). Default **on** for waiting/errored.
- Prefer OS notify APIs (`osascript` / `notify-send` / similar) with terminal bell as fallback, not herdr-style pane injection as the only path.

## Out of scope

- Full in-manager toast/action system (#77)
- Cross-session messaging (#62)
- Mobile / remote push

## Acceptance

- Start an agent, leave it on a permission prompt: one notification fires when status becomes `waiting`.
- Leaving it blocked does not re-fire every poll.
- Toggle off in Settings silences them.
- Documented in usage/settings.

## Priority

High for product wedge / herdr parity on the "who needs me" story. Ship before larger notification redesign.

redeye1011:
## What changed

- New `internal/notify` package: one notification per status transition.
  When the drawing terminal is Ghostty or cmux (TERM_PROGRAM=ghostty,
  CMUX_WORKSPACE_ID, or TERM=xterm-ghostty — the one marker plain SSH
  carries) it fires OSC 777 via the existing termseq tmux passthrough: a
  native notification attributed to the right window/workspace, which also
  reaches the user when the manager runs on a remote host over SSH, since
  the escape rides the terminal connection. Otherwise macOS posts via
  osascript (argv form, no quoting pitfalls) and Linux via notify-send;
  the terminal bell is the floor, so headless and WSL setups still get an
  audible cue.
- The poller's existing transition branch (`newStatus != sess.Status` in
  refreshOnce) drives it: waiting/errored fire exactly once per
  transition — never per poll tick — and finished only with the opt-in.
- Settings gains `notifications` (default on) and `notify on finish`
  (default off), persisted as store settings like the other toggles.

## Why

Closes #226

Supersedes the contrib/ scripts from #269: built-in means brew and
install.sh users get it with the binary, nothing rots between releases,
and detection reads the in-process transition instead of locking state.db
and the status files into a public contract.

Focus: deliberately no focus gate. The poll cannot tell whether the
manager is being watched, and gating on the selected row would suppress
the ping exactly when the manager sits unfocused in another workspace —
the case notifications exist for. Happy to revisit once a real focus
signal exists.

## How it was verified

- Unit tests: every delivery branch (Ghostty OSC 777 on both platforms,
  including the SSH case where only TERM=xterm-ghostty crosses; osascript;
  notify-send; bell floor), payload sanitize, poller transition gating
  (waiting/errored fire, finished opt-in, setting silences), settings
  persistence, and an integration test driving a real pane through
  refreshOnce into waiting — fires once, steady state never re-fires.
- Live: dev build against real sessions on macOS (cmux) and on a remote
  Ubuntu host over SSH — notifications arrived on the local machine,
  attributed to the correct workspace, exactly once per transition.
- `gofmt -l .`, `go vet ./...`, `go test -race ./...`, `go build ./...`
  clean on macOS arm64 and Linux x86_64.

- [x] `go vet ./...` passes

## Summary by CodeRabbit

* **New Features**
  * Added desktop and terminal notifications for waiting and errored status transitions.
  * Added optional notifications when tasks finish, with duplicate alerts suppressed.
  * Added persistent notification controls in Settings, including keyboard navigation.
  * Added platform-specific delivery through desktop notifications, terminal alerts, and bell fallback.
  * Notification content is sanitized, and unsupported delivery methods fail silently.

* **Documentation**
  * Updated usage documentation to describe notification behavior, delivery options, and settings.


YoanWai:
Thanks for the quick turnaround on this! Two things before I review: CI is red on `TestRefreshNotifiesWaitingTransitionOnce`, it passes locally for me so likely the session is already `waiting` before the test installs the recorder on a slow runner, worth pinning the pre-state. And there are 3 open CodeRabbit comments.

redeye1011:
Thanks — all three addressed in `4a50c09`:

- **Bounded delivery**: external notifiers now run under `exec.CommandContext` with a 2s timeout, so a wedged `osascript`/`notify-send` costs one delivery and falls through to the bell instead of hanging.
- **Off the critical path**: `notifyTransition` dispatches delivery in a goroutine, so `refreshOnce` never waits on an external process while holding `runMu`.
- **Comment trim**: the restating comments in `keys.go` / `settings.go` / `notify_test.go` are gone.

And the CI flake: you called it — the test could observe a transition before the recorder was installed. It now pins the pre-state (stored status forced to `idle` before the recorder goes in) and awaits the async fire with a deadline instead of asserting synchronously. 20/20 under `-race` locally; full suite green on macOS arm64 and Linux x86_64.

redeye1011:
One deliberate omission, flagged as a follow-up rather than added here: delivery currently leaves no trace — a missed ping (wedged notify-send, terminal without OSC 777 support, tmux eating the escape) gives the user no way to tell a fire-and-lose from never-fired. A small delivery journal inside the notify module (last N attempts: session, status, path taken, error) behind a debug surface would close that without touching the Notify interface. Kept out of this PR to keep the slice thin — happy to send it as a follow-up PR if you think it earns its place.

YoanWai:
Keeping this out of the PR was the right call. I don’t think we need the delivery journal yet, so no need to spend time on a follow up for now.

If people start reporting missed notifications that we can’t explain, the small recent history you described sounds like the right thing to reach for. Thanks for flagging it.

PR #280, opened 2026-08-12
feat(ui): label every filter over the sessions list instead of in the header

YoanWai:
The archive named itself twice, once as a header badge and once over the list, while the attention filter showed only in the header and hidden empty groups showed nowhere at all. A filter narrows the list, so the list is where it belongs.

Every active filter now paints a badge over the entries, next to the key that lifts it:

```
  ARCHIVED   t back to active
  ATTENTION   w show all
  HIDE EMPTY   e show empty

   root                 ● 1
 ● Anna  finished · grok · 8d ago
```

The header keeps the count and a plain scope word (`6 agents · archived`), and the badges follow the rail's existing fit rules: padded block when there is room, bare badges on a tighter rail, dropped only when entries would have no room left under them.

Verified in the running TUI on an isolated tmux socket at 200 and 90 columns, with each filter alone and all three together; `gofmt`, `go vet` and `go test ./...` are clean.

## Summary by CodeRabbit

* **Enhancements**
  * Filter indicators for archived items, attention status, and hidden empty groups now appear together in the navigation rail.
  * Filter badges adapt to available space and remain visible on shorter layouts.
  * Reset and “show all” actions are displayed directly with the relevant filter badges.
  * Filter badges no longer repeat in the list header, keeping the interface clearer.

* **Documentation**
  * Updated status-filter guidance to explain how badges, archived views, counts, and list contents reflect active filters.

### DIFF
--- a/docs/usage.md
+++ b/docs/usage.md
@@ -186,7 +186,7 @@ Each session's tmux pane is polled (default every 2s) to derive a status:
 | `idle` | Nothing running |
 | `dead` | The tmux session is gone |
 
-`w` narrows the list to sessions that need attention (`waiting`, `finished`, `errored`). Press again to show every status. The header badge and session counts follow the filter; folds open so matches are not hidden.
+`w` narrows the list to sessions that need attention (`waiting`, `finished`, `errored`). Press again to show every status. An `ATTENTION` badge sits over the list with the key that clears it, and the session counts follow the filter; folds open so matches are not hidden. The archived view (`t`) and hidden empty groups (`e`) label themselves the same way, so the list always names every filter it is under.
 
 ![the session tree, with a waiting agent's permission prompt in the preview](screenshot-sessions.png)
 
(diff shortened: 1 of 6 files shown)


YoanWai:
@coderabbitai review

YoanWai:
Both applied in df2069f.

- **docs/usage.md:189** — the "always" claim is gone. The note now says the badges take whatever room the rail has: padded on a tall terminal, tight against the entries on a short one, and yielding to the entries once the list is down to its last rows.
- **listview_test.go** — the test now reads the rail's first painted lines in order, so a badge that drifted below the entries fails, and each line has to carry its own key (`ARCHIVED` beside `t back to active`, and so on). It also asserts the header keeps the plain `· archived` scope word alongside rejecting the badge labels.

YoanWai:
Applied in 11309d8: the comment restated the assertions under it, so it is gone rather than reworded.

On the earlier `test` failure: `TestFocusWatchReportsCursor` died on `tmux load-buffer ...: server exited unexpectedly`, a tmux-lifecycle flake in the focus watcher, nowhere near the rail rendering here. The rerun of the same commit passed.

PR #281, opened 2026-08-12
test: anchor the tmux test servers so they survive between tests

YoanWai:
`TestFocusWatchReportsCursor` has failed four CI runs since August 1 (Aug 1, 2, 4, and PR 280's first run), always the same way: `tmux load-buffer ...: exit status 1: server exited unexpectedly`.

The mechanism is the one 3675f4a first diagnosed. Every test kills its tmux sessions in cleanup, and a server whose last session dies begins an exit-empty shutdown. A fresh session created on that socket inside the shutdown window lands on the dying server and goes down with it, so the next command on the new session reports `server exited unexpectedly`. 3675f4a isolated one source of dying servers (the backoff test's transient one), but every kill-last-session boundary on the shared socket still bounces the server, and `ReportsCursor` sits right after such a boundary. A PID check shows the bounce directly, even on a fast machine where the failure window is too narrow to hit:

```
-- without anchor --
pid before=54097 after=54250 bounced=yes
-- with anchor --
pid before=54275 after=54275 bounced=no
```

The fix: `TestMain` in both tmux-driving packages (`internal/ui`, `internal/tmux`) creates an `anchor` session right after the initial kill-server. The server always holds at least one session, so the shutdown window never opens; the final kill-server still tears everything down. Session names the driver uses all carry the `am_` prefix, so the anchor is invisible to every existing test.

Both suites pass on an isolated socket, and the full `go test ./...` is green.

## Summary by CodeRabbit

* **Tests**
  * Improved test environment stability by keeping the terminal session available while individual tests run.
  * Ensured test cleanup continues to remove temporary sessions and shut down the test environment afterward.

### DIFF
--- a/internal/tmux/tmux_test.go
+++ b/internal/tmux/tmux_test.go
@@ -16,8 +16,13 @@ import (
 const testSocket = "amtmuxtest"
 
 // TestMain kills any leftover test server so each run starts and ends clean.
+// The anchor session then holds the server up for the whole run: tests kill
+// their sessions in cleanup, and a server whose last session dies begins an
+// exit-empty shutdown that takes the next test's fresh session down with it
+// ("server exited unexpectedly").
 func TestMain(m *testing.M) {
 	tmuxCmd("kill-server").Run()
+	tmuxCmd("new-session", "-d", "-s", "anchor").Run()
 	code := m.Run()
 	tmuxCmd("kill-server").Run()
 	os.Exit(code)
--- a/internal/tmux/tmux_test.go
+++ b/internal/tmux/tmux_test.go
@@ -1,6 +1,7 @@
 package tmux
 
 import (
+	"fmt"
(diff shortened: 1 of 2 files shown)


YoanWai:
@coderabbitai review

YoanWai:
@coderabbitai review

YoanWai:
@coderabbitai full review

YoanWai:
Applied in e0751e5, with one deliberate carve-out.

- Anchor creation now aborts the run loudly when tmux is present and `new-session` fails: a run without the anchor would pass or flake on luck. Machines without tmux still reach `m.Run()` so each test skips through its own `requireTmux`.
- The `kill-server` calls stay unchecked, before and after: with no server up they fail as their normal case (every fresh run, every machine without tmux), and failing the run on teardown would mask the tests' own result.

Issue #292, opened 2026-08-13
TestRefreshNotifiesWaitingTransitionOnce times out on CI


YoanWai:
The notification test flaked on the first PR that ran after #276 merged ([job](https://github.com/YoanWai/agent-manager/actions/runs/31709948160/job/94479979718)):

```
--- FAIL: TestRefreshNotifiesWaitingTransitionOnce (5.40s)
    notify_test.go:174: want 1 notifications, got []
```

Re-running the same commit passed, and the test passes locally under `-race -count=5`.

`waitForCalls` gives the delivery 5 seconds (`notify_test.go:38`). The path it waits on is the full one: a hooks status file, a refresh, a store write, then the notification off the refresh path, with a real tmux pane behind it. On a loaded runner that is close enough to the deadline to lose, and the failure is a timeout rather than a wrong result, so the deadline is what to look at.

Either give the wait a deadline sized for CI, or drive the delivery deterministically instead of polling for it.

YoanWai:
Measured on an unchanged `main`, running the CI command four times in one job ([run](https://github.com/YoanWai/agent-manager/actions/runs/31728969686)):

```
=== iteration 1   ok    internal/ui  63.894s
=== iteration 2   --- FAIL: TestRefreshNotifiesWaitingTransitionOnce (5.69s)
=== iteration 3   ok    internal/ui  61.472s
=== iteration 4   ok    internal/ui  61.518s
```

So roughly one run in four, on the branch with nothing added to it. Some more of the shape of it, from chasing a red check on #293:

- The test alone, five times over, passes every time in about 0.3s.
- The whole `internal/ui` package under `-race` without coverage passes.
- It only turns up under the CI command, `-race` with `-covermode=atomic`, which roughly doubles how long everything takes.
- Adding two store reads and a `t.Logf` around the refresh made it pass on a branch that had failed four times running, so the window it misses is a narrow one.

`refreshOnce` takes `runMu` and blocks, so the manual refresh in the test always runs; the 5s in `waitForCalls` is spent waiting for a notification that was never going to come, rather than one that was late. That points at the transition being consumed before the refresh the test drives, by a poll it does not control, rather than at delivery being slow.

tonydzi:
hi — this is Mycroft, Anton Dziatkovskii's synthetic cofounder (AI). the analysis below is from reading the code at e1ae486, not from a live repro, so the trigger is labeled hypothesis where it is one.

[...]

PR #295, opened 2026-08-13
probe: is TestRefreshNotifiesWaitingTransitionOnce failing on a clean main?

YoanWai:
Throwaway. A whitespace-only change on top of `origin/main` so CI runs the suite with nothing else in it, to tell #292 apart from the branch it was noticed on. Closing as soon as the job reports.


YoanWai:
Answered its question: a whitespace-only change on top of main passed, so #292 is a flake rather than a break on main. Closing.

PR #296, opened 2026-08-13
probe: diagnose the notify test failure

YoanWai:
Throwaway, closing as soon as CI reports. Logs the stored status around the refresh so #292's failure can be read instead of guessed.


YoanWai:
Answered: the same branch passed here, so the notify failure on #293 is #292's flake rather than anything in the change. Closing.

PR #297, opened 2026-08-13
probe: notify test failure rate and status trace

YoanWai:
Throwaway; closing once CI reports.


YoanWai:
Answered; the flake rate is measured on #292. Closing.

PR #298, opened 2026-08-13
probe: notify flake rate on clean main

YoanWai:
Throwaway; four coverage+race runs of internal/ui on an unchanged main to measure #292's failure rate. Closing once it reports.


YoanWai:
Measured: 1 failure in 4 coverage+race runs of internal/ui on an unchanged main. Recorded on #292.

PR #309, opened 2026-08-14
fix(status): keep a session working while its background shells run

### LINKED ISSUE #308: Background shells and agents notify "finished" while they are still running
### What happens

A turn that leaves background work running reports `finished`, so the desktop notification fires while the session is still busy. When that work completes the agent is re-invoked, that turn ends too, and a second notification fires. One request, two pings.

### Where it comes from

Claude Code fires `Stop` when the main agent stops responding, whatever is still running underneath. The pane says what is left, and that is the signal the status engine already reads for background agents:

- `✻ Worked for 3s · 1 shell still running`
- `✻ Cooked for 4s · 2 shells still running`
- `✻ Waiting for 1 background agent to finish`

`busy_line` matches the third shape only, so a turn holding a background shell resolves through `turn_end` as a finished turn.

Background agents have a second hole. `isBusy` requires the wait line to be the newest content, and transient banners render under it (`Plugins updated: 7 plugins · Run /reload-plugins to apply`). With one of those below the wait line the pane matches nothing at all, and the hook's `finished` passes straight through.

### Reproduction

Claude Code 2.1.232, isolated tmux socket, a settings file logging every hook event next to the status write.

One prompt that spawns a background subagent:

```
UserPromptSubmit          working
PreToolUse   Agent        working
PostToolUse  Agent        working
Stop                      finished   <- subagent still running
PreToolUse   Bash  AGENT  working    <- the subagent's own tool call
SubagentStop AGENT
UserPromptSubmit          working    <- re-invoked by the completion
Stop                      finished
```

One prompt that starts a `run_in_background` shell, status file sampled every 1.5s:

```
14:44:57 working
14:45:02 finished   <- sleep 40 still running
...
14:45:40 working    <- shell exited, agent re-invoked
14:45:41 finished
```

Both windows derive `finished` from the pane as well, so the status the list shows and the notification that fires agree with each other and disagree with the session.

YoanWai:
## What changed

`busy_line` for `claude` now matches the shell shape of a turn-end summary as well as the background-agent wait line, so a turn holding `run_in_background` work keeps reading as `working`:

```
✻ Cooked for 4s · 2 shells still running
✻ Waiting for 1 background agent to finish
```

A config still carrying the agents-only pattern takes the current one through the tool merge, the way a legacy Codex working rule already does; a hand-edited pattern is left as its author wrote it.

`isBusy` now certifies a busy line by asking whether a turn ended below it, rather than whether the line is the newest content. A tool with no `turn_end` keeps its busy line until it stops drawing it, which is how hermes reports its delegations.

## Why

Claude Code fires `Stop` the moment the main agent stops responding, whatever is still running underneath. A turn holding a background shell reported `finished`, so the desktop notification landed while the work was still going. The shell then exited, the agent was re-invoked, that turn ended too, and a single request pinged twice.

Background agents had the hole from the other side. A banner rendered under the wait line (`Plugins updated: 7 plugins · Run /reload-plugins to apply`) read as a newer turn, so the wait was no longer the newest content, the pane matched nothing, and the hook's `finished` passed straight through.

Closes #308

## How it was verified

Claude Code 2.1.232 driven in an isolated tmux socket, with a settings file logging every hook event next to the status write. Three runs captured: one background shell, two background shells, one background subagent. Panes captured every 1.5s through each run.

Every captured pane replayed through the engine, loaded from the real config directory:

| newest marker in the pane | before | after |
| --- | --- | --- |
| `✻ Worked for 3s · 1 shell still running` | `finished` | `working` |
| `✻ Cooked for 4s · 2 shells still running`, banner under it | unmatched, so the hook's `finished` stood | `working` |
| `✻ Waiting for 1 background agent to finish`, banner under it | unmatched, so the hook's `finished` stood | `working` |
| `✻ Sautéed for 43s`, agents drained | `finished` | `finished` |

The hook event log for one background subagent, showing the two turn ends behind the two pings:

```
UserPromptSubmit          working
PreToolUse   Agent        working
PostToolUse  Agent        working
Stop                      finished   <- subagent still running
PreToolUse   Bash  AGENT  working    <- the subagent's own tool call
SubagentStop AGENT
UserPromptSubmit          working    <- re-invoked by the completion
Stop                      finished
```

- [x] `go vet ./...` passes

## Summary by CodeRabbit

- **Bug Fixes**
  - Improved activity detection for background shells and agents.
  - Continued work is now correctly shown as active until a later turn completes.
  - Prevented plugin banners and other extra output from incorrectly marking work as finished.
  - Existing custom status configurations remain unchanged during upgrades.

- **Documentation**
  - Clarified how background shells and agents are represented in status detection.


YoanWai:
@coderabbitai full review

YoanWai:
@coderabbitai review

YoanWai:
@coderabbitai review

YoanWai:
@coderabbitai review

YoanWai:
@coderabbitai review

PR #312, opened 2026-08-14
fix(ui): align focused-pane selection with display cells

OkunaRei:
## What changed

Focused-pane selection now uses terminal display columns and complete Unicode grapheme clusters. A drag that lands inside a wide glyph, or on an NFD combining mark or a keycap sequence, copies the whole cluster.

`--help` from the original patch lives in a follow-up so this PR stays on selection.

## Why

Mouse coordinates are cells. CJK glyphs occupy two of them, so treating a cell as a rune index selected text to the right of the pointer. The same walk has to keep combining marks and keycaps whole, because `DecodeSequenceInString` returns each ASCII byte immediately.

## How it was verified

- [x] `go vet ./...` passes
- [x] `go test -race ./...` passes (`env -u TMUX TMUX_TMPDIR=/tmp/am312`)

Tests cover CJK, ZWJ emoji, NFD `é`, and `#️⃣` on both drag and word select.

### DIFF
--- a/internal/ui/focussel.go
+++ b/internal/ui/focussel.go
@@ -316,17 +316,16 @@ type graphemeSpan struct {
 
 func graphemeSpans(line string) []graphemeSpan {
 	var spans []graphemeSpan
-	for offset, cell, state := 0, 0, byte(0); offset < len(line); {
-		text, width, size, nextState := ansi.GraphemeWidth.DecodeSequenceInString(line[offset:], state, nil)
-		if size <= 0 {
+	for offset, cell := 0, 0; offset < len(line); {
+		text, width := ansi.FirstGraphemeCluster(line[offset:], ansi.GraphemeWidth)
+		if text == "" {
 			break
 		}
 		spans = append(spans, graphemeSpan{
-			start: offset, end: offset + size, startCell: cell, endCell: cell + width, text: text,
+			start: offset, end: offset + len(text), startCell: cell, endCell: cell + width, text: text,
 		})
-		offset += size
+		offset += len(text)
 		cell += width
-		state = nextState
 	}
 	return spans
 }
--- a/internal/ui/focussel.go
+++ b/internal/ui/focussel.go
@@ -299,7 +299,7 @@ func (m *Model) expandSelection() {
 	switch m.sel.granule {
 	case selectLine:
 		m.sel.anchorCol = 0
-		m.sel.headCol = len(line)
+		m.sel.headCol = ansi.StringWidth(string(line))
 	case selectWord:
 		start, end := wordBounds(line, m.sel.anchorCol)
 		m.sel.anchorCol, m.sel.headCol = start, end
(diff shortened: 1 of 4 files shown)


OkunaRei:
Addressed the review findings in 9d54d80:

- selection now maps terminal cells to complete Unicode grapheme clusters, including a drag endpoint inside a wide glyph and the ZWJ sequence `👩‍💻`
- inactive selection rendering skips the display-width pass
- `printHelp` propagates output errors
- subprocess coverage exercises both `--help` and `-h` through `main()` without a TTY

Validation: `gofmt -l .`, `go vet ./...`, and `env -u TMUX TMUX_TMPDIR=/tmp/am312 go test ./...`.

OkunaRei:
@coderabbitai review

internal/ui/focussel.go
```diff
@@ -295,35 +295,80 @@ func (m *Model) expandSelection() {
...
+type graphemeSpan struct {
+	start, end         int
+	startCell, endCell int
+	text               string
+}
+
+func graphemeSpans(line string) []graphemeSpan {
+	var spans []graphemeSpan
+	for offset, cell, state := 0, 0, byte(0); offset < len(line); {
+		text, width, size, nextState := ansi.GraphemeWidth.DecodeSequenceInString(line[offset:], state, nil)
```
YoanWai: **[bug]** `graphemeSpans` walks stripped text with `ansi.GraphemeWidth.DecodeSequenceInString`. That decoder returns each ASCII byte immediately (`c > US && c < DEL`) and never consumes a following combining mark, variation selector, or keycap. `ansi.StringWidth` / `FirstGraphemeCluster` do cluster those, so cell counts and byte slices disagree. Verified at this head: double-click on NFD `"e\u0301 x"` copies `"e"` (NFC `"é x"` copies `"é"`); double-click on `"#\ufe0f\u20e3"` copies `"#"`; a one-cell drag of NFD `é` also drops the acute. CJK and `👩‍💻` start on non-ASCII, so the new tests do not catch this. The same hole is why a drag endpoint "inside" those clusters cannot keep them whole.

**Suggestion:** Iterate with `ansi.FirstGraphemeCluster(line[offset:], ansi.GraphemeWidth)` (or `graphemes.FromString` plus the same width method `StringWidth` uses). The line is already stripped, so the ANSI decoder is not needed. Add regressions for NFD `"e\u0301 x"` and `"#\ufe0f\u20e3"` on both drag and double-click.

main.go
```diff
@@ -60,6 +61,13 @@ func main() {
 	update.SetBuildSource(buildSource)
 
 	if len(os.Args) > 1 {
+		if os.Args[1] == "--help" || os.Args[1] == "-h" {
```
YoanWai: **[suggestion]** CONTRIBUTING asks that a PR stay on one thing. The focused-pane cell/grapheme fix and root `--help`/`-h` are independent: two commits, two packages, two failure modes. They can land separately without blocking each other.

**Suggestion:** Split `printHelp` and the `os.Args[1] == "--help"` / `"-h"` branch into their own PR. The selection change can merge on its own once the grapheme walk is fixed.

YoanWai:
Rebased onto current main, switched the grapheme walk to `ansi.FirstGraphemeCluster` (NFD and keycap sequences were still splitting), and moved `--help` into a follow-up so this PR is selection-only.

YoanWai:
Follow-up for `--help`: #322.

YoanWai:
Shipped in [v0.30.0](https://github.com/YoanWai/agent-manager/releases/tag/v0.30.0). Thanks, @OkunaRei: a drag in a focused pane now copies the cluster it covered, CJK, emoji and combining marks included, instead of running off to the right of the pointer.

PR #315, opened 2026-08-14
feat(ui): hold a placeholder name until a new agent renames itself

### LINKED ISSUE #158: Hold a placeholder name until a new agent renames itself
When quick prompting a new agent, or creating one with a built-in prompt, the agent name always gets renamed after creation. Currently the randomly generated name briefly appears before the rename completes, which is visually jarring.

We should show a placeholder (e.g. a loading skeleton or "..." state) instead of the randomly generated name while renaming is in progress. Only if renaming fails should we stop the loading state and fall back to showing the generated name.

YoanWai:
## What changed

A spawn that asks its agent to name itself launches under a generated name, and the manager now holds that name back until the agent answers. Every reading of the session shows the same stand-in while the wait is on: the rail row, the group roster beside it, the detail head, and the quick bar target, all through one `displayName`.

The wait ends the moment the rename lands, and the generated name takes over when no answer can come: the pane died, or a minute of grace covering the boot, the directive, and the command the agent runs has passed. Sessions the user named, shells, and terminals show their name from the first frame, because nothing was ever asked to rename them.

## Why

Quick prompts and built-in prompts always rename the agent right after creation, so the row flashed a throwaway name like `claude-ab12` and swapped it for the real one seconds later. The row now settles on one name.

Closes #158

## How it was verified

```
$ gofmt -l .
$ go vet ./...
$ env -u TMUX TMUX_TMPDIR=/tmp/am158 go test ./...
ok  	github.com/YoanWai/agent-manager	1.613s
...
ok  	github.com/YoanWai/agent-manager/internal/ui	143.923s
ok  	github.com/YoanWai/agent-manager/internal/update	(cached)
ok  	github.com/YoanWai/agent-manager/tools/badges	(cached)
```

`gofmt -l .` printed nothing and `go vet ./...` printed nothing. The suite drives real tmux sessions on the isolated `/tmp/am158` socket.

```
$ env -u TMUX TMUX_TMPDIR=/tmp/am158 go test -race ./...
ok  	github.com/YoanWai/agent-manager	3.645s
...
ok  	github.com/YoanWai/agent-manager/internal/ui	160.900s
ok  	github.com/YoanWai/agent-manager/internal/update	2.567s
ok  	github.com/YoanWai/agent-manager/tools/badges	2.718s
```

Tests covering the behaviour:

- `TestEveryReadingOfASessionStandsInForAnAwaitedName` spawns an auto-named session into a group and reads the roster, the detail head and the quick bar. Run against the code before this fix it failed on all three, each printing `claude-ab12` while the rail row printed the stand-in.
- `TestSessionRowStandsInForAnAwaitedName` walks the five states of a row: waiting on the agent, rename landed, pane dead, grace expired, and a spawn that was never asked to rename. Each case also asserts the wait drops what it was holding once it is over.
- `TestSpawnAwaitsARenameOnlyWhenItAsksForOne` covers the auto-named and user-named spawns.
- `TestTerminalRowShowsItsNameImmediately` covers a shell.

Re-run after rebasing onto `ac909b6`: `gofmt -l .`, `go vet ./...` and `go test -race ./...` are all green at `0228146`, the current head.

- [x] `go vet ./...` passes

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved session naming while newly spawned sessions await an agent-selected name.
  * Added consistent placeholder and name handling across session lists, details, rosters, and quick access views.
  * Preserved immediate display of names for custom-named sessions and shell sessions.
  * Cleaned up pending naming state when sessions are renamed, expire, or are deleted.

* **Tests**
  * Added coverage for pending-name display, expiration, deletion, and custom naming behavior.

### DIFF
--- a/internal/ui/form.go
+++ b/internal/ui/form.go
@@ -618,7 +618,7 @@ func (m *Model) spawnSession(toolName, name, dir, group, prompt string, autoName
 		agentSessionID = uuid.NewString()
 		base += " " + tool.SessionIDFlag + " " + agentSessionID
 	}
-	return m.launchNewSession(store.Session{
+	if err := m.launchNewSession(store.Session{
 		ID:    id,
 		Name:  name,
 		Tool:  toolName,
@@ -633,7 +633,18 @@ func (m *Model) spawnSession(toolName, name, dir, group, prompt string, autoName
 		PendingInputs:  pendingInputs,
 	}, tool, base, launchOptions{
 		rollbackWorktree: worktreeRepo != "",
-	})
+	}); err != nil {
+		return err
+	}
+	// The directive went out with the launch, so the row waits for the name
+	// the agent picks instead of showing the one generated for it.
+	if autoNamed {
+		if m.awaitedRenames == nil {
+			m.awaitedRenames = map[string]string{}
+		}
+		m.awaitedRenames[id] = name
+	}
+	return nil
 }
 
 func withPrompt(tool config.Tool, command, prompt string) string {
--- a/internal/ui/form_test.go
+++ b/internal/ui/form_test.go
@@ -389,6 +389,29 @@ func TestLaunchPromptInjectsDirectiveOnlyForAutoNamedWithPrompt(t *testing.T) {
 	}
 }
 
+// Only a spawn that hands over the rename directive has a name to wait for;
+// one the user named itself is already at its final name.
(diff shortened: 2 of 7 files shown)


YoanWai:
@coderabbitai full review

YoanWai:
@coderabbitai full review

PR #316, opened 2026-08-14
fix(ui): expand a pane row's tabs into the cells it painted

### LINKED ISSUE #288: Literal tabs in pane previews corrupt the TUI layout
## What happened

1. Open a managed terminal session in the agent-manager repository.
2. Run:

   ```sh
   env -u TMUX TMUX_TMPDIR=/tmp/amtest go test ./...
   ```

3. Select that terminal session in the manager.

Expected: the captured pane is rendered within the preview column and the rest of the TUI remains stable.

Actual: `go test` result rows cause the frame to overflow and wrap. The preview, footer, and computer-stat rows are then painted at incorrect screen positions, making the TUI appear corrupted. The underlying tmux pane remains intact.

## Cause

`go test` emits literal tab bytes (`0x09`) between the result, package, and duration fields. `previewLine` intentionally preserves tabs, while `ansi.StringWidth` does not account for their terminal tab-stop expansion. Padding and truncation therefore produce a row that is logically within the preview width but physically wider when iTerm expands the tabs. The overflow breaks Bubble Tea renderer cursor bookkeeping.

Captured example:

```text
ok  \tgithub.com/YoanWai/agent-manager/internal/ui\t73.163s
```

The preview should expand tabs to spaces using pane-relative tab stops before measuring, truncating, and padding the row.

## agent-manager version

`dev` at `2bb986fbee1d02bd56624ac2a6c025bd4bf90737`

## Environment

- macOS
- iTerm2
- tmux 3.7b
- Not tool-specific; reproduced in a managed terminal session

## Additional evidence

`tmux capture-pane` confirms the pane buffer is clean and contains literal `0x09` bytes. Clearing the pane removes the triggering rows and lets the manager repaint normally.

YoanWai:
## What changed

`paneExact` now turns each captured row into the cells the pane actually painted: a tab byte becomes the spaces it covers, out to the next eight-column stop, and a tab close to the right edge stops on the row's last cell, which is where tmux leaves the cursor. Rows reach every consumer already expanded, so the frame's measure and clip, the caret splice and the mouse selection all count one set of columns, the same ones tmux reports the cursor and the mouse in.

## Why

`go test` writes literal tabs between its result, package and duration fields. `ansi.StringWidth` gives a tab no columns, so a row that measured as fitting the preview came out wider once the host terminal expanded it: the frame overflowed and wrapped, and the preview, footer and computer-stat rows were painted at the wrong screen positions. Focus mode counted the same rows the same way, so on a tabbed row the caret and the selection highlight landed a tab's worth of padding away from the text they belonged to, and a drag copied the wrong characters.

Closes #288

## How it was verified

tmux's own behavior at the right edge, on the isolated socket (tmux 3.7b, 62-column pane):

```
env -u TMUX TMUX_TMPDIR=/tmp/am288 tmux -f /dev/null new-session -d -s p2 -x 62 -y 10 "printf '%s\tXY' '<58 a>'; sleep 60"
env -u TMUX TMUX_TMPDIR=/tmp/am288 tmux -f /dev/null display-message -p -t p2 '#{cursor_x},#{cursor_y}'
env -u TMUX TMUX_TMPDIR=/tmp/am288 tmux -f /dev/null capture-pane -p -t p2 | od -c
```

58 columns then a tab leaves the cursor at 61, X paints on column 61 and Y wraps to the next row, while `capture-pane` still serializes the row as 58 a's, `\t`, `X`. The expansion pads that tab by 3, which is what the pane shows.

End to end through the real render path: a temporary test created a session, sent the `go test` row into its pane, polled `capture-pane` until the row came back with its `0x09` intact, then rendered `m.View()` at 100x30. Every one of the 30 frame rows measured exactly 100 cells and the preview row read `ok      github.com/YoanWai/agent-manager/internal/ui    73.163s`. With the expansion removed the same check reports the tab surviving into the painted rows.

```
cd /Users/yoan/Desktop/projects/agent-manager-worktrees/issue-288-preview-tabs
gofmt -l .
go vet ./...
env -u TMUX TMUX_TMPDIR=/tmp/am288 go test ./...
env -u TMUX TMUX_TMPDIR=/tmp/am288 go test -race ./...
```

`gofmt -l .` printed nothing, `go vet ./...` was clean, and both test runs came back `ok` for every package (`internal/ui` 80.3s, 132.6s under `-race`).

- [x] `go vet ./...` passes

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved pane rendering and text selection when content contains tab characters.
  * Ensured tabs align consistently to eight-column stops across previews, cursors, and mouse selections.
  * Preserved pane width and blank-space rendering at right edges and beyond narrow frames.

### DIFF
--- a/internal/ui/focusscroll_test.go
+++ b/internal/ui/focusscroll_test.go
@@ -195,7 +195,7 @@ func TestFocusScrollRecapturesAfterPreviewResize(t *testing.T) {
 	}
 	updated, _ = m.Update(recapture())
 	m = updated.(*Model)
-	if got, want := len(paneExact(m.preview, m.previewPaneHeight())), m.previewPaneHeight(); got != want {
+	if got, want := len(paneExact(m.preview, m.previewPaneHeight(), m.previewPaneWidth())), m.previewPaneHeight(); got != want {
 		t.Fatalf("scroll frame has %d rows, want %d", got, want)
 	}
 	if !strings.Contains(m.preview, "history-line-") {
@@ -238,7 +238,7 @@ func TestFocusScrollKeepsDeepHistoryFrame(t *testing.T) {
 	}
 	updated, _ := m.Update(msg)
 	m = updated.(*Model)
-	if got, want := len(paneExact(m.preview, m.previewPaneHeight())), m.previewPaneHeight(); got != want {
+	if got, want := len(paneExact(m.preview, m.previewPaneHeight(), m.previewPaneWidth())), m.previewPaneHeight(); got != want {
 		t.Fatalf("deep frame has %d rows, want %d", got, want)
 	}
 	if !strings.Contains(m.preview, "history-line-") {
--- a/internal/ui/focussel.go
+++ b/internal/ui/focussel.go
@@ -108,7 +108,7 @@ func (m *Model) paneCell(x, y int) (row, col int, ok bool) {
 // same slice the renderer paints, so selection indices line up with what
 // is on screen.
 func (m *Model) paneTextLines() []string {
-	rows := paneExact(m.preview, m.pane.box.height)
+	rows := paneExact(m.preview, m.pane.box.height, m.pane.box.width)
 	out := make([]string, len(rows))
 	for i, row := range rows {
 		out[i] = ansi.Strip(previewDangerSeqs.ReplaceAllString(row, ""))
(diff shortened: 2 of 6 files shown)


YoanWai:
@coderabbitai full review

YoanWai:
@coderabbitai full review

YoanWai:
@coderabbitai full review

PR #324, opened 2026-08-16
feat: project setup and run scripts

roshaans:
## Why

A worktree is a fresh checkout. Everything git does not track is missing from it — installed dependencies, an `.env`, generated files — so an agent started into one spends its first turn on errors that are not the code's, and there is no way to start the project's dev server or test watcher from the manager at all.

This lets a repository declare both, in a committed `.agent-manager/settings.toml`:

```toml
setup = """
npm ci
cp ../myapp/.env .
"""

[run.dev]
command = "npm run dev -- --port $AGENT_MANAGER_PORT"
default = true

[run.test]
command = "npm test -- --watch"
```

Repository settings rather than user settings on purpose: what it takes to bootstrap a project is the same for everyone working on it, so it belongs next to the code rather than in one person's `~/.config/agent-manager/config.toml`.

Everything is opt-in — a repository with no settings file behaves exactly as it does today.

## What it does

**`setup`** runs in the session's own pane, in the new worktree, before the agent — so it is watched rather than waited on behind a spinner. On success the agent starts exactly as it would have. On failure it does not start at all, and the pane says which exit code it got and leaves a shell in the worktree to fix it from (`R` restarts once fixed). Only worktree spawns run it; a session pointed at a checkout you prepared by hand is left alone.

Settings are read from the worktree, so they version with the branch. A settings file written but not yet committed is not in the checkout — which is exactly the state a project is in the first time anyone tries this — so that case falls back to the repository the worktree branched from, and the first attempt works instead of silently doing nothing.

**`p`** runs a named script in a terminal tab beside the session under the cursor, in that session's directory. One script, or one marked `default = true`, starts straight away; several with no default open a picker, because the wrong pick here starts a server rather than doing nothing. Each is an ordinary shell session in the list — same scrollback, same `x`/`v`/`a`/`d`. It is named for both the script and the session it runs beside, so `dev` in two worktrees stays tellable apart.

**Ports.** Running five agents at once is not much use if their dev servers all want port 3000. Every worktree gets a block of ten ports, with `$AGENT_MANAGER_PORT` exported into setup scripts, run scripts and the agent's own pane. The block is derived from the worktree's directory name rather than handed out by a counter, so it survives restarts of both the session and the manager — a worktree keeps the address you bookmarked for as long as it keeps its name — and a block already listening is skipped, so two names that collide can still both run. `port_base` moves the range; it defaults to 3100, above where a hand-started `npm run dev` usually lands.

## One change outside the feature

`envCommand` now exports its variables on their own lines instead of assigning them inline. The inline form only covers a single simple command: `K=v foo && bar` leaves `bar` without the variable, and a `$K` inside `foo` expands before the assignment takes effect. A run script is a shell snippet, so this had to change for the port to reach it. It also carries the existing variables into the shell the launch script `exec`s afterwards, so a pane that drops to a prompt keeps them. This matches what `writeLaunchScript` already does for `COLORFGBG`.

## Notes

- New `internal/project` package, 20 tests, plus 12 in `internal/ui`. The port test asserts the variable actually reaches a live pane rather than just the struct; the setup test asserts the script actually runs in a real worktree.
- The setup wrapper deliberately does not `exec`. `writeLaunchScript` ends with its own `exec $SHELL`, which is what leaves a usable prompt once the command finishes; exec'ing here would take that away, so a failed setup would kill the pane instead of leaving somewhere to fix it from.
- This repo now dogfoods it: `.agent-manager/settings.toml` declares `go mod download && go build ./...` as setup and `test`/`check`/`tui` as run scripts.
- Docs in `docs/project-settings.md`, linked from the README and the keys table.
- `gofmt`, `go vet ./...` and `go test ./...` all clean.

Happy to split the `envCommand` change into its own commit, rename the key, or move the settings file location if you'd prefer something else.

## Summary by CodeRabbit

* **New Features**
  * Added project settings for setup commands, run scripts, defaults, and per-worktree port allocation.
  * Added the `p` key to run configured project scripts, with selection when multiple scripts are available.
  * Added project-specific setup and environment handling for worktree sessions.
  * Added browser access for running local servers and double-`Esc` navigation back to the manager.
  * Added safeguards for invalid settings and failed setup commands.

* **Documentation**
  * Added project settings documentation and linked it from the README.
  * Documented the new run-script keyboard shortcut.

### DIFF
--- a/.agent-manager/settings.toml
+++ b/.agent-manager/settings.toml
@@ -1,0 +1,22 @@
+# How agent-manager bootstraps and runs its own worktrees.
+# Reference: docs/project-settings.md
+
+# A fresh worktree has no module cache warmed and no binary built. Doing it
+# here means the agent's first `go test` is not also the first download.
+setup = """
+go mod download
+go build ./...
+"""
+
+[run.test]
+command = "go test ./..."
+description = "the whole suite"
+default = true
+
+[run.check]
+command = "gofmt -l . && go vet ./... && go test ./..."
+description = "what CI runs"
+
+[run.tui]
+command = "go run ."
+description = "this manager, from source"
--- a/.agent-manager/settings.toml
+++ b/.agent-manager/settings.toml
@@ -1,22 +1,35 @@
 # How agent-manager bootstraps and runs its own worktrees.
 # Reference: docs/project-settings.md
 
-# A fresh worktree has no module cache warmed and no binary built. Doing it
-# here means the agent's first `go test` is not also the first download.
-setup = """
+# A fresh worktree has nothing warmed: no module cache, no build. Doing it
+# here means the agent's first `go test` is not also the first download, and
+# a worktree that cannot build says so before an agent starts working in it.
+setup = '''
 go mod download
 go build ./...
-"""
+'''
+
+# Testing a worktree's build means running this manager inside this manager.
+# HOME is redirected at a directory inside the worktree because config.Dir()
+# resolves under it: without that, a half-finished build would be reading and
+# writing the real ~/.config/agent-manager/state.db — your actual sessions —
+# and a bug in the thing under test would land on the sessions you were using
+# to test it. The dev instance therefore starts empty, every time.
+#
+# O attaches to this pane rather than opening a browser, since a TUI has
+# nothing to serve.
+[run.tui]
+command = '''
+mkdir -p .am-dev-home
+go build -o .am-dev . && HOME="$PWD/.am-dev-home" exec ./.am-dev
+'''
+description = "this manager, built from this worktree, on its own state"
+default = true
 
 [run.test]
 command = "go test ./..."
 description = "the whole suite"
-default = true
 
 [run.check]
 command = "gofmt -l . && go vet ./... && go test ./..."
 description = "what CI runs"
-
-[run.tui]
-command = "go run ."
-description = "this manager, from source"
--- a/.gitignore
+++ b/.gitignore
@@ -16,3 +16,7 @@ docs/badges/contributors.png
 *.p12
 id_rsa
 id_ed25519
+
+# dev builds of this manager, made by [run.tui]
+/.am-dev
+/.am-dev-home/
--- a/README.md
+++ b/README.md
@@ -33,7 +33,7 @@ The tools you use alongside agents live in the same list: `T` opens a plain shel
 
 Not here yet: cost tracking, mouse-driven navigation, and agents that can talk to each other.
 
-**Jump to:** [Install](#install) · [Usage](#usage) · [Keys](docs/usage.md#keys) · [Diff review](docs/usage.md#diff-review) · [Configuration](docs/configuration.md) · [Docs site](https://agent-manager.dev/docs/)
+**Jump to:** [Install](#install) · [Usage](#usage) · [Keys](docs/usage.md#keys) · [Diff review](docs/usage.md#diff-review) · [Configuration](docs/configuration.md) · [Project settings](docs/project-settings.md) · [Docs site](https://agent-manager.dev/docs/)
 
 ## Supported tools
 
--- a/docs/project-settings.md
+++ b/docs/project-settings.md
@@ -9,6 +9,20 @@ in one person's `~/.config/agent-manager/config.toml`.
 Everything here is opt-in. A repository without the file behaves exactly as
 it did before.
 
+## Creating it
+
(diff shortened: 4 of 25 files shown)


roshaans:
Thanks — all five addressed in 30e4cb7. Details, since two of them changed behaviour:

**Settings discovery no longer climbs out of the repository.** This was the right catch and the most serious one. `Load` now takes the repository root and stops there; `Setup` and every `Run.Command` reach a shell, so a file planted in a shared parent directory was someone else's code running as the user. An empty root reads the given directory alone, which is the safe reading for a caller that cannot say where the repository begins, and callers derive the bound from `git rev-parse --show-toplevel` — for a worktree that is the worktree itself, so neither the committed nor the uncommitted-fallback read can escape.

One wrinkle worth flagging: paths are compared symlink-resolved. Without that, a repository reached through a symlink — `/tmp` on macOS is one — never matches its own root and the walk runs past it. My first attempt resolved the walk but not the bound, and the new `TestLoadWithoutARootReadsOnlyThatDirectory` caught it climbing anyway.

**`port_base` is validated at load** rather than clamped at use, so an out-of-range value fails as the setting that caused it rather than as a server that will not start. Boundary tests cover `-1`, `0`, `64537`, `70000`, and both accepted edges.

**The whole block is probed**, not just its base — a project deriving a second port from `$AGENT_MANAGER_PORT` could otherwise be handed a block whose upper half was already taken. Test occupies `taken+1`.

**`form_test.go` now covers both new branches** beside the code they exercise: the fallback to uncommitted source-repository settings (asserting the settings really are absent from the worktree, so the test cannot silently stop covering the fallback), and the load error that must discard the created worktree and leave no session behind.

**Comment restating `Dir` and `File` removed.**

`gofmt`, `go vet ./...` and `go test ./...` are clean. Docs updated for the repository bound and the port-range rules.

roshaans:
Pushed the TUI half onto this branch rather than opening a second PR — it is the same feature and the review here already covers the code it touches. Three commits added on top of the review fixes:

- `feat: create project settings from the TUI`
- `feat: author project settings from the group form, export $PORT`
- `fix: keep the settings path when there is no editor to open it with`

**What they add.** Nothing in the manager could write a `settings.toml`, so `p` in an unconfigured repository reported an error naming a path you then had to leave the manager to create — discoverable exactly where it was unusable.

The new-group form now asks for a **setup** and a **run** command alongside name, path and worktree, since setting up a project is when you know what it takes to bootstrap it. Answer either and the file is written at the repository root; answer neither and nothing is written, because a group is not necessarily a project you asked to configure. A repository that already has settings shows them read-only with `e` to open the real file — rewriting hand-edited TOML from a form would have to preserve comments, key order and every block the form has no field for, which is an editor, and the manager can open yours instead. `e` opens the file from the run picker too, and a settings file with no run scripts now opens rather than dead-ending on an error.

`p` in a repository with no settings offers to scaffold one: `↵` writes a fully commented starter and opens it. Commented out throughout, so creating it cannot change how anything already works, and an existing file is never overwritten.

**`$PORT`.** The port is now exported under that name as well as `$AGENT_MANAGER_PORT`. `PORT` is what node, rails and flask already read, so an unmodified `npm run dev` in five worktrees serves on five ports with the project unchanged — the difference between per-worktree ports being a feature and being something you have to wire up. Neither name is set for a project with no settings file.

**Writing TOML safely.** Values are written as literal strings, which have no escape processing, so a backslash, a `$` or a quote in a shell command survives the write/read round trip; a value carrying the literal delimiter falls back to an escaped basic string. `TestWriteRoundTripsThroughLoad` covers both over nine inputs. Worth flagging one subtlety it caught: TOML trims the newline *after* a `'''` opening delimiter but not the one before the closing delimiter, so a symmetric-looking implementation round-trips a trailing `\n` into the value.

Three further bugs came from tests rather than review: `startRun` built its env by hand and so never received `$PORT`; the group form kept the previous project's prefilled values after the path was retargeted; and opening the settings file replaced the message naming that file with the editor's own failure, so a machine with no editor learned only that nothing happened.

`gofmt`, `go vet ./...` and `go test ./...` are clean. #325 is closed.

YoanWai:
Sorry, I completely missed this before you closed it. You put a lot of work into this and I should have replied.

What made you decide to close it? Was it the size of the change, the automated review, or something else? I don’t want to guess whether you still want to pursue it, but I do want to understand what happened.

And thank you for all the work you put into it.

PR #325, opened 2026-08-16
feat: author project settings from the TUI

roshaans:
**Builds on #324 — please read that one first.** Cross-fork PRs can only target a branch in this repo, so this is opened against `main` and therefore currently shows #324's commits too. Once #324 merges, the diff here collapses to just the three commits below. Happy to squash the two together instead if you'd prefer one change.

The commits that are new here:
- `feat: create project settings from the TUI`
- `feat: author project settings from the group form, export $PORT`
- `fix: keep the settings path when there is no editor to open it with`

## Why

#324 added `.agent-manager/settings.toml` but nothing in the manager could write one. Pressing `p` in an unconfigured repository reported an error naming a path you then had to leave the manager to create — the feature was discoverable exactly where it was unusable.

## Setting up the project is when you know how to run it

The new-group form now asks, alongside name, path and worktree:

```
  name      backend
  parent    ~
  path      ~/Code/myapp
  worktree  ◂ on ▸
  setup     npm ci
  run       npm run dev
```

Answer either and `.agent-manager/settings.toml` is written at the repository root. Answer neither and nothing is written — a group is not necessarily a project you asked to configure.

A repository that already has settings shows them read-only, with `e` to open the real file. Rewriting hand-edited TOML from a form would have to preserve comments, key order and every block the form has no field for, which is an editor; the manager can open yours instead. `e` opens the file from the run picker too, and a settings file with no run scripts now opens rather than dead-ending on an error.

`p` in a repository with no settings offers to scaffold one — `↵` writes a fully commented starter at the repository root and opens it. Commented out throughout, so creating it cannot change how anything already works, and an existing file is never overwritten.

## $PORT

The port is now exported as `$PORT` as well as `$AGENT_MANAGER_PORT`. `PORT` is what node, rails, flask and most of the ecosystem already read, so an unmodified `npm run dev` in five worktrees serves on five ports with the project unchanged — the difference between per-worktree ports being a feature and being a thing you have to wire up.

Neither name is set for a project with no settings file, so a project that never opted in does not have `PORT` changed underneath it.

## Writing TOML safely

Values are written as TOML *literal* strings, which have no escape processing, so a backslash, a `$` or a quote in a shell command survives being written and read back exactly as typed. A value carrying the literal delimiter falls back to an escaped basic string. `TestWriteRoundTripsThroughLoad` covers both paths over nine inputs (`sed -i "s/a/b/"`, `echo it's fine`, embedded `'''`, tabs, multi-line).

One subtlety worth flagging: TOML trims the newline *after* a `'''` opening delimiter but not the one before the closing delimiter, so only the leading newline is written — a symmetric-looking implementation round-trips a trailing `\n` into the value. The round-trip test caught that.

## Notes

- Three bugs were caught by tests rather than by review: `startRun` built its env by hand and so never received `$PORT`; the group form kept the previous project's prefilled values after the path was retargeted; and opening the settings file replaced the message naming that file with the editor's own failure, so a machine with no editor learned only that nothing happened (the nix sandbox has no editor on `PATH`, which is how that surfaced).
- `openEditor`'s launch is split out as `openInEditor` so a single file can be opened, not just a directory. Editors take either.
- `gofmt`, `go vet ./...` and `go test ./...` clean.
- Docs updated in `docs/project-settings.md`.

### DIFF
--- a/.agent-manager/settings.toml
+++ b/.agent-manager/settings.toml
@@ -1,0 +1,22 @@
+# How agent-manager bootstraps and runs its own worktrees.
+# Reference: docs/project-settings.md
+
+# A fresh worktree has no module cache warmed and no binary built. Doing it
+# here means the agent's first `go test` is not also the first download.
+setup = """
+go mod download
+go build ./...
+"""
+
+[run.test]
+command = "go test ./..."
+description = "the whole suite"
+default = true
+
+[run.check]
+command = "gofmt -l . && go vet ./... && go test ./..."
+description = "what CI runs"
+
+[run.tui]
+command = "go run ."
+description = "this manager, from source"
--- a/README.md
+++ b/README.md
@@ -33,7 +33,7 @@ The tools you use alongside agents live in the same list: `T` opens a plain shel
 
 Not here yet: cost tracking, mouse-driven navigation, and agents that can talk to each other.
 
-**Jump to:** [Install](#install) · [Usage](#usage) · [Keys](docs/usage.md#keys) · [Diff review](docs/usage.md#diff-review) · [Configuration](docs/configuration.md) · [Docs site](https://agent-manager.dev/docs/)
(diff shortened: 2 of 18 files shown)


roshaans:
Folded into #324 — these belong together, and the review there already covers the same code. #324 now carries the TUI authoring commits as well.

Issue #329, opened 2026-08-16
chore: migrate the TUI to Bubble Tea v2


YoanWai:
The TUI is pinned to Bubble Tea v1.3.10, plus Bubbles v1.0.0 and Lip Gloss v1.1.0. v2 of all three was tagged 2026-02-24, five months before this repo's first commit (`72f6972`, 2026-07-15), and `go.mod` has never been touched for these deps since. Nothing forced v1; it was the scaffold's default and it stuck.

## Why it matters now

v1 cannot see `ctrl+shift+<letter>`. Its key table has no such entries, and it never asks the terminal for the kitty/modifyOtherKeys protocol, so `ctrl+o` and `ctrl+shift+o` arrive as the same byte, `0x0F`. Verified by feeding a v1.3.10 program `\x1b[27;5;111~`: it reports `KeyRunes "27;5;111~"` and types the sequence as text.

That caps three things:

- #326 (user-configurable keybindings) can only ever offer keys v1 can parse.
- Moving the in-session editor key off `ctrl+o`, which Claude Code, Codex and others already use, has no good target in focus mode.
- Any future shortcut needing Shift, or key-release events, is unreachable.

v2 has `View.KeyboardEnhancements` and `KeyPressMsg{Code, Mod}` with `ModShift`, which is exactly the missing capability. It also ships `RequestBackgroundColor` / `BackgroundColorMsg` (bears on #268) and a real `View.Cursor` (bears on #266).

## What the migration touches

Counted across the tree:

| Area | Volume | Nature |
| --- | --- | --- |
| `tea.KeyMsg` refs | 270, of which 247 in tests | `KeyMsg` becomes an interface; press handling moves to `KeyPressMsg` |
| `tea.Key<Const>` refs | 326, of which 253 in tests | `msg.Type` becomes `msg.Code` + `msg.Mod` |
| Key literals in tests | ~250 sites, 53 distinct forms | `tea.KeyMsg{Type: tea.KeyEnter}` and friends; regular enough for a scripted rewrite plus one helper |
| Mouse | 80 refs over `focusscroll.go`, `focussel.go`, `split.go`, `model.go` | `MouseMsg` becomes an interface, events split into Click/Release/Wheel/Motion, buttons renamed |
| `View() string` | 1 impl (`internal/ui/view.go:15`) | returns `tea.View`; alt-screen and mouse mode become fields on it |
[...]

Issue #326, opened 2026-08-16
feat: user-configurable keybindings


YoanWai:
Every key the manager reacts to is hardcoded, in the TUI keymap and in the tmux server bindings alike. When one of them collides with a key the user or the agent inside a session already owns, there is no way to move it.

`EnsureBindings` (`internal/tmux/tmux.go:283`) installs four root-table binds on the `agentmgr` socket:

```
bind-key -n C-q  if-shell -F '#{m:am_*,#{session_name}}' detach-client 'send-keys C-q'
bind-key -n C-\  ...
bind-key -n C-r  ... review
bind-key -n C-o  ... editor
```

Inside an `am_*` session the key is consumed and the agent never sees it. Claude Code binds Ctrl+O to its own "show more lines" action, so attaching to a Claude session (#290 shipped the Ctrl+O editor bind) costs the user a key their agent uses. Ctrl+R is Claude's history search on the same footing. Unbinding by hand does not hold: `EnsureBindings` reinstalls on every manager start and on every session create (`internal/ui/model.go:802`, `internal/tmux/tmux.go:244`).

#265 was the same shape from the other side, a user whose tmux prefix was `C-\` losing every documented exit at once. PR #302 made the live exits discoverable, which helps that case without giving anyone a way to choose their own keys.

## Proposal

A `[keybindings]` table in `config.toml` mapping named actions to keys, scoped by where the key is read:

```toml
[keybindings.session]   # tmux root table, inside am_* sessions
editor = "none"          # skip the bind entirely; ctrl+o reaches the agent
review = "ctrl+g"
detach = ["ctrl+q", "f9"]

[keybindings.list]      # manager TUI
editor = "e"

[keybindings.review]    # fullscreen diff review
```

- Actions carry names (`editor`, `review`, `detach`, and the list actions), so a key is data rather than a `case` arm.
- `"none"` disables an action's binding, which is what closes this report: the key passes through to whatever runs inside the session.
- Defaults live in the same table, so an empty config behaves exactly as today.
- A list value binds several keys to one action, covering #265's "bind a function key too".

Vim is the reference worth copying: mode-scoped map tables, keys written by name (`<C-o>`), `<Nop>` to disable, and `:verbose map <C-o>` to find who owns a key.

## Surfaces to feed from the same table

- tmux binds, `internal/tmux/tmux.go:283`
- session footer text, `internal/tmux/tmux.go:262`
- focus title, `internal/ui/listview.go:691`
[...]

PR #331, opened 2026-08-16
fix(tmux): move the in-session editor key to F3

YoanWai:
## What changed

The in-session editor key moves from `Ctrl+O` to `F3`, in a full-screen attach and in focus mode alike. `EnsureBindings` also unbinds `C-o`, so a tmux server that was already running when the manager updated drops the old binding instead of carrying it until the next reboot. The footer, the focus rule, the key map and the docs all name `F3` now.

## Why

`Ctrl+O` belongs to the programs running inside a session. Claude Code binds it to `app:toggleTranscript`, Gemini CLI to `TOGGLE_COPY_MODE` and `EXPAND_PASTE`, readline to `operate-and-get-next`, and `nano` to write-out. Holding it in tmux's root key table took it from all of them, and a user attaching to a Claude session lost a key their agent uses.

`F3` was picked by checking every candidate against the agents we support and against the terminals people run us in:

| Key | Why not |
| --- | --- |
| `F1` | The VS Code integrated terminal keeps it: `workbench.action.showCommands` is in the default `commandsToSkipShell` |
| `F2` | opencode binds `f2` and `shift+f2` to model cycling |
| `F4`, `F6`–`F9`, `F12` | Gemini CLI default bindings |
| `F5`, `F10`, `F11` | VS Code keeps them for the debugger; `F10` also opens the GTK menu and `F11` is fullscreen nearly everywhere |

`F3` is free in Claude Code, Codex, Gemini CLI, opencode and pi, passes through the VS Code integrated terminal, and carries no default binding in tmux (0 F-keys in the root table), Windows Terminal (only `F11` and `Alt+F4`) or GNOME Terminal. Konsole binds it to Find Next, which is rebindable there.

A `ctrl+shift+<letter>` was the other candidate and cannot work: Bubble Tea v1.3.10 has no `ctrl+shift+<letter>` key and never requests the kitty/modifyOtherKeys protocol, so focus mode receives `0x0F` for both `ctrl+o` and `ctrl+shift+o`. #329 tracks the v2 migration that lifts that ceiling, and #326 tracks making all of this configurable.

## How it was verified

- `env -u TMUX TMUX_TMPDIR=/tmp/amtest go test -race ./...` — all packages pass, including the new `TestEnsureBindingsMovesTheEditorKeyToF3` and the `ctrl-o` case in `TestFocusKeyCommand`.
- Drove a real attach on an isolated socket with a pty client: sending `0x0F` left the request marker empty and the client attached, so the key reached the pane; sending `\x1bOR` and `\x1b[13~` both set `@am_request=editor` and detached the client.
- `tmux -L … show-options -t am_… -v status-right` reads `agent-manager · Ctrl+r = review · F3 = editor · Ctrl+q = back`.
- Recorded both binaries through vhs against a fake `HOME` and an isolated `TMUX_TMPDIR` for the screenshots below.

- [x] `go vet ./...` passes
- [x] `go build ./...` passes

## Visual evidence

**Before:**

Focus mode, `main` at 51b76a8:

focus mode before

Full-screen attach, `main` at 51b76a8:

attach footer before

**After:**

Focus mode: the rule and the footer both read `f3 editor`.

focus mode after

Full-screen attach: the session footer reads `F3 = editor`.

attach footer after

## Known limitations

`F3` is free in every agent CLI and terminal checked, with two exceptions worth naming: Konsole binds it to Find Next, where it is rebindable in Konsole's own settings, and a Mac laptop sends media keys on the function row unless "Use F1, F2, etc. keys as standard function keys" is on, so it is `Fn+F3` there. That is true of any F-key.

No single fixed key is free everywhere: every `ctrl+letter` is claimed by an agent CLI or readline, each F-key by some agent or terminal, `alt` is unreliable on macOS defaults, and `ctrl+shift+<letter>` cannot reach focus mode at all under Bubble Tea v1. #326 is where this stops being a fixed choice.

## Summary by CodeRabbit

* **Changes**
  * The shortcut for opening a session’s directory in the editor is now **F3** instead of **Ctrl+O**.
  * Updated session help, footers, key references, and usage documentation to reflect the new shortcut.
  * **Ctrl+O** remains available for applications running inside sessions.
  * Existing session bindings are updated automatically to use **F3**.


YoanWai:
@coderabbitai review

YoanWai:
@coderabbitai full review

PR #341, opened 2026-08-18
fix(ui): preserve Unicode and colors in pane selection

creatorrr:
## What

- encode focused-pane selections as UTF-16LE before sending them to Windows `clip.exe` under WSL
- preserve the pane's ANSI styling before and after agent-manager's selection overlay
- cover both paths with focused regression tests, including non-BMP Unicode and non-WSL clipboard writers

## Why

Agent-manager owns click-drag selection while a child pane is focused. On WSL, v0.30 sent that UTF-8 text directly to `clip.exe`, which interpreted the bytes through the Windows OEM code page. Box-drawing characters such as `─` consequently pasted as `ΓöÇ`.

The selection renderer also rebuilt every selected row from plain text. Selecting even one span therefore removed the agent's colors from the rest of that row.

## Visual evidence

This is a CLI-only submission, so I cannot attach the local screenshots directly. The deterministic text reproduction is:

```text
before copy: ────────────────────
before paste: ΓöÇΓöÇΓöÇΓöÇΓöÇΓöÇ

after copy:  ────────────────────
after paste: ────────────────────
```

For selection rendering, a row containing red `red`, a selected separator, and blue `blue` previously lost both surrounding ANSI colors. The regression test now asserts that both color sequences survive the overlay.

## Verified

- `gofmt -l .`
- `go vet ./...`
- `env -u TMUX TMUX_TMPDIR=/tmp/amup go test -race ./...`
- `go build ./...` from a fresh local clone of this branch

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved clipboard text handling on WSL for correct Unicode support, including emoji and other surrogate-pair characters.
  * Preserved existing text encoding behavior on other platforms.
  * Fixed selection highlighting so surrounding terminal colors and styling remain intact.
  * Preserved wide characters and graphemes when rendering selected text.

* **Tests**
  * Added coverage for WSL Unicode clipboard output and styled selection rendering.

### DIFF
--- a/internal/clipboard/clipboard.go
+++ b/internal/clipboard/clipboard.go
@@ -3,8 +3,10 @@
 package clipboard
 
 import (
+	"bytes"
 	"context"
 	"encoding/base64"
+	"encoding/binary"
 	"errors"
 	"fmt"
 	"io/fs"
@@ -14,6 +16,7 @@ import (
 	"runtime"
 	"strings"
 	"time"
+	"unicode/utf16"
 
 	"github.com/YoanWai/agent-manager/internal/termseq"
 	"github.com/charmbracelet/x/ansi"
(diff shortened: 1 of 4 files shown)


YoanWai:
Thanks @creatorrr! The WSL UTF-16LE encoding matches the confirmed working recipe from microsoft/WSL#11047, and the color-preserving overlay is a real improvement to focus mode.

I pushed one commit on top before merging: the splice truncated at the raw selection columns while the selected span snapped outward to whole graphemes, so a wide grapheme (CJK, emoji) under a selection edge rendered on both sides and duplicated. e933bd3 truncates at the snapped byte edges instead, with a regression test for the emoji case.

This ships in the next release.

Issue #213, opened 2026-08-04
Renderer broken (flickers) since >= 0.18


creatorrr:
### What happened

screen recording:

https://github.com/user-attachments/assets/6e2b241c-82fe-4fff-ad65-a12327dab1ba

### agent-manager version

v0.18+

### Operating system

Windows (WSL2)

### tmux version

tmux next-3.7

### Which agent CLI

Claude Code

### Relevant config

```toml

```

YoanWai:
thanks for the report, on it

YoanWai:
Root cause found: since v0.18 every preview row is wrapped in invisible bidi format characters (U+200E/U+2066/U+2069) to contain RTL pane text. Windows Terminal under conpty renders those zero-width characters as a full cell (microsoft/terminal#8667), so rows overflow, the terminal scrolls, and the whole frame repaints on every preview tick.

Fix in #214: the wrapper is now added only to rows that actually contain RTL text, so LTR-only rows match pre-0.18 output byte for byte.

Could you confirm on your setup? `git clone -b worktree-issue-213-preview-flicker https://github.com/YoanWai/agent-manager && cd agent-manager && go build -o am-test . && ./am-test`

creatorrr:
Confirmed — built `worktree-issue-213-preview-flicker` (bde1307) and verified it on the machine from the report. Your root cause is correct, and I can now back it with direct measurements from every layer of my stack. A few findings below that matter for #214 before it merges.

## Ground truth: who charges cells for the marks

Measured on this machine (Windows 11 build 26200, WezTerm → `wsl.exe` → WSL2 Ubuntu 22.04 → tmux next-3.7):

| layer | U+200E LRM | U+2066 LRI | U+2069 PDI | `LRM+LRI+"hello"+PDI` |
|---|---|---|---|---|
| Windows console buffer (ConPTY/conhost), measured via `[Console]::CursorLeft` | **1** | **1** | **1** | **8** |
| tmux next-3.7 cursor advance, measured via `#{cursor_x}` | 0 | 0 | 0 | 5 |
| `x/ansi` v0.11.7 / `uniseg` v0.4.7 / `lipgloss` v1.1.0 | 0 | 0 | 0 | 5 |
| `go-runewidth` v0.0.23 | 0 | 1 | 1 | 7 |

[...]

PR #354, opened 2026-08-19
fix(ui): paint RTL rows the same on every host

YoanWai:
## What

Frame rows carrying Hebrew or Arabic are painted as the cells they occupy, on every host. The LRM and directional isolate that wrapped those rows on iTerm2, and the `TERM_PROGRAM` / `LC_TERMINAL` sniffing that switched them on, are gone, along with `AGENT_MANAGER_RTL_PIN`.

## Why

iTerm2 3.6.11 drops the marks before they reach its grid, so the host they shipped for never saw them. Printed three rows into a session, one bare, one with a leading LRM, one inside the isolate: its own `text` property reads all three back byte for byte identical, with no format characters left. Screenshots of a fresh window with right-to-left support enabled show a Hebrew-leading row holding its column with the marks and without them.

The hosts that do act on the marks are the ones the marks broke: WezTerm with `bidi_enabled` reorders the row until it no longer matches what was painted (#286), and Windows Terminal under conpty charges a cell for a format character and overflows the row (#213).

So the render path now hands every terminal the same frame. A terminal that runs its own bidirectional layout still reorders it, which its own right-to-left setting controls, and `docs/configuration.md` says so.

## Verified

- `go build ./...`, `go vet ./internal/ui`, `go test ./internal/ui/...` green (209s, isolated `TMUX_TMPDIR`).
- Live: the built binary on an isolated tmux socket with a throwaway `HOME`, one terminal session focused, a line mixing Hebrew inside an English sentence wide enough to wrap. `capture-pane` shows the pane rows with the same text and column as before and no `U+200E` / `U+2066` / `U+2069` anywhere in the frame.
- `previewLine` keeps its column width, its text, and its SGR handling for RTL rows; `TestPreviewLine` asserts the row carries no bidi controls.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved right-to-left text rendering so Hebrew and Arabic rows display consistently within their occupied cells across supported hosts.
  * Removed unwanted bidirectional control markers from preview rows while preserving text and layout width.
  * Updated terminal behavior to maintain consistent column ordering.

* **Documentation**
  * Clarified that terminal-native bidirectional layout settings, including those in iTerm2 and WezTerm, must be disabled for correct display.


YoanWai:
@coderabbitai full review

PR #364, opened 2026-08-21
fix(ui): keep a just-spawned session on screen

YoanWai:
A session spawned while a poll pass was in flight vanished from the tree until the next pass came round. On a busy machine with a large fleet that is a visible blink; recording the docs demos is where it turned up.

## What was happening

`refreshOnce` reads the session list at the top of its pass, then spends the rest of it in tmux and `ps` calls. The UI applied the result wholesale, so a list assembled before the spawn erased the row that had just been added optimistically.

## The fix

The UI carries over the rows this run launched that the poll had not looked for yet. The first poll to list the store after a launch is the authority on it, whether it reports the row or its absence.

Launch time is stamped by the model rather than read from `CreatedAt`: building the tmux window takes 40-90ms before the store write, so a pass listing inside that window misses a row whose own timestamp already predates it.

## Second commit

Review caught three ways the first one was wrong, each reproduced through the real key path:

- a session **deleted** inside the window came back, with no tmux window behind it
- a session **archived** inside the window came back onto the live tree wearing the dead status the kill had just given it
- **focus** tested the poll's own list rather than the merged one, so spawning into focus and catching a stale poll dropped out of the session it had just entered

Both lifecycle paths now forget the launch as they send the session away, and focus reads the list the UI actually adopts.

## Verification

Four tests, each written before its fix and watched fail. `gofmt` clean, `go vet` clean, full suite green on an isolated tmux socket.

### DIFF
--- a/internal/ui/lifecycle.go
+++ b/internal/ui/lifecycle.go
@@ -617,6 +617,7 @@ func (m *Model) applyConfirmedArchived(archived bool) error {
 		if err := m.store.SetArchived(sess.ID, archived); err != nil {
 			return err
 		}
+		m.forgetLaunch(sess.ID)
 	}
 	return nil
 }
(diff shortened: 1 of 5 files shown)


YoanWai:
@coderabbitai review

PR #365, opened 2026-08-21
feat(ui): name a waiting row by the prompt it was given

YoanWai:
A session spawned without a name holds a placeholder until the agent answers the rename directive, which takes as long as a boot plus a turn. Five agents launched in a burst all wore the same `…` for that whole minute, so the tree could not say which row was which task.

## What the row wears now

The prompt it was given, capped at what the narrowest rail affords beside a starting row's state, tool and age:

```
add cursor …   starting up · claude · 0s ago
backfill th…   starting up · codex  · 0s ago
rate limit …   starting up · grok   · 0s ago
```

It is correct whether the rename lands in two seconds, in sixty, or never, so the name arriving is the only thing that changes. Past the one-minute grace the row still falls back to its generated `claude-ab12`, exactly as before.

## Why the prompt is carried rather than read back

`store.Session.LaunchPrompt` is the **decorated** prompt: it carries the rename directive and the coordination note that were sent to the agent with it. Painting that would put "Run this exact shell command once, replacing `<name>`…" in the row. The raw prompt rides beside the generated name in the record that already tracks the wait, which is run-scoped and needs no schema change.

## Second commit

Review caught a real defect. A pasted image reaches the agent as the path it was written to, so a prompt opening with a screenshot named the row after a temp directory — and every image-first spawn produced the same string, which is worse than the placeholder it replaced. The pictures now drop out of the preview and the words stay.

## Cap

12 columns. Measured against the rail: at 20 the row lost its tool and its age, and at 16 and 14 the age rendered half-truncated. 12 is the width where the row keeps the shape every other row has.

## Verification

Six tests, the primary one watched fail first. The naive implementation was written deliberately to confirm the guard tests catch a leaked directive and an unflattened prompt — they do. `gofmt` clean, `go vet` clean, full suite green on an isolated tmux socket.

### DIFF
--- a/internal/clipboard/clipboard.go
+++ b/internal/clipboard/clipboard.go
@@ -58,6 +58,13 @@ func pastesDir() string {
 	return filepath.Join(os.TempDir(), "agent-manager-pastes")
 }
 
+// IsPastePath reports whether a word is one of the image paths a pasted
+// picture becomes on its way to an agent, so a reader showing the prompt
+// back can leave the pictures out of it.
+func IsPastePath(word string) bool {
+	return filepath.Dir(word) == pastesDir()
+}
+
 // StaleAfter is how long a pasted image stays on disk before a sweep can
 // take it: long enough for an agent to reopen an image from an earlier
 // session, short enough that screenshots do not pile up in temp.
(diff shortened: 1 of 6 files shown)


YoanWai:
@coderabbitai review

PR #374, opened 2026-08-22
fix(ui): never shrink a pane, crop the preview instead

### LINKED ISSUE #369: Resizing the preview clears a Codex session's whole scrollback
Every genuine height change we push to a managed pane makes a running Codex clear that pane's entire scrollback and replay its transcript. Output that predates Codex is destroyed and does not come back. Resizing the terminal window, or attaching and detaching a session, is enough to trigger it.

## Measured

`codex-cli 0.147.0`, inline mode (`alternate_on` is 0), an isolated tmux server. 400 unique marker lines printed by the shell, then Codex started and taken past its update prompt:

```
before:  markers=401  history_size=413
after a height-only resize 100x30 to 100x44:
         markers=0    history_size=1
```

## Why it happens

`Driver.Resize` runs `resize-window -x -y` (`internal/tmux/tmux.go:554-560`). It is called from `internal/ui/model.go:999` when the preview box relayouts and from `internal/ui/model.go:1323` on attach and detach.

On the Codex side at tag `rust-v0.147.0`, `resize_reflow.rs:350` sets `should_rebuild_transcript = reflow_needed || height_changed`, so a height change alone qualifies, and `clear_terminal_for_resize_replay` (`:267`) calls `clear_scrollback_and_visible_screen_ansi()`.

## Direction

Same-size resizes are already inert on both sides: we skip an unchanged `WindowSizeMsg` (`internal/ui/model.go:1011-1016`) and skip any session whose cached geometry already matches, and Codex compares against its own `last_known_screen_size`. So the lever is reducing genuine height deltas, by debouncing relayout or keeping managed panes out of preview box resizing.

Codex configuration is not the lever. The clear is not gated on `tui.terminal_resize_reflow_max_rows`, so setting that key does not prevent it, and setting it to 0 costs scrollback in its own right: at 100x30 it leaves 686 non-blank rows against 1249 by default.

## Reproducing without spending tokens

`codex resume <rollout-id>` replays a stored transcript with no model call. Dismiss the update offer by checking the selection line reads `2. Skip` before pressing Enter, so no upgrade is triggered. The wipe needs Codex's interface actually running; a Codex still sitting on its update prompt does not show it.

YoanWai:
Fixes #369.

## The bug

Every height shrink the manager pushed through `resize-window` made a running Codex (`codex-cli` 0.147.0, inline mode) clear the pane's **entire scrollback** and replay its transcript. Output that predated Codex was destroyed for good. Shrinks fired from transient chrome relayouts, from a shorter terminal, and at startup for sessions adopted from a previous run.

Measured per tool with the same harness (400 seeded marker lines, one real turn, height-only `resize-window` 100x30→44→30): **codex wipes in both directions** (markers 400→0, history 583→175). claude and opencode run on the alt screen and are untouched; gemini keeps history across resizes (it clears pre-existing scrollback during its first turn on its own, which no manager behavior can affect).

## The fix

`resizeSessions` pins width changes and height growth only; a box that lost rows leaves the pane tall:

- `paneWindow` crops the render at the last painted row, so a taller pane shows its content with no dead band: short output stays top-anchored, a not-yet-painted pane keeps its blank rows hit-testable, and the caret's row always stays on screen.
- The crop start feeds `paneRowOffset`, so the caret cell and mouse forwarding map through the same window the renderer paints.
- Focus-mode scroll regions anchor to the pane's bottom through the capture itself (`-S -offset -E -`, cut in Go), so a taller pane scrolls its real history without knowing the pane height.
- Sessions adopted from a previous run seed their true geometry from one `list-windows` call and are never shrunk to fit a box they already exceed.
- A window created this run takes one exact pin (its launch size comes from pre-selection geometry, and its scrollback is still empty, so the pin is free).

Attach and detach still reflow: the width genuinely changes there, and a width reflow rebuilds the transcript in any terminal. The one cosmetic cost, an alt-screen TUI cropped at the top while its pane is taller than the box, heals on the next attach.

## Verified

- Full test suite green on an isolated tmux server, including new coverage: chrome shrink leaves the pane tall and grows it back past the pin, a height-only terminal shrink keeps the pane, adopted taller panes survive the first pass, `paneWindow` crop and caret math, and the bottom-anchored scroll window.
- End to end under the built manager driving a real replayed Codex transcript (`codex resume`, no tokens spent), against the shipped release as control:

| trigger | shipped v0.25 | this branch |
|---|---|---|
| terminal height resize | history 205→174, pre-Codex lines destroyed | history byte-identical |
| quick bar open/close | stable (already held by #342) | stable |
| selection flips | stable | stable |
| terminal shrink | pane re-pinned, replay | pane kept tall, render crops, history intact |

### DIFF
--- a/internal/tmux/tmux.go
+++ b/internal/tmux/tmux.go
@@ -652,6 +652,36 @@ func noServer(out string) bool {
 		strings.Contains(out, "error connecting to")
 }
 
+// WindowSizes returns every managed session's window width×height in a
+// single tmux call, so sessions adopted from a previous run resize from
+// their real geometry rather than from nothing.
+func (d *Driver) WindowSizes() (map[string][2]int, error) {
+	out, err := exec.Command(d.bin, d.args("list-windows", "-a", "-F", "#{session_name} #{window_width} #{window_height}")...).CombinedOutput()
+	if err != nil {
+		if noServer(string(out)) {
+			return map[string][2]int{}, nil
+		}
+		return nil, fmt.Errorf("tmux list-windows: %w: %s", err, strings.TrimSpace(string(out)))
+	}
+	sizes := map[string][2]int{}
+	for _, line := range strings.Split(strings.TrimSpace(string(out)), "\n") {
+		fields := strings.Fields(line)
+		if len(fields) != 3 || !strings.HasPrefix(fields[0], prefix) {
+			continue
+		}
+		id := strings.TrimPrefix(fields[0], prefix)
+		if _, taken := sizes[id]; taken {
+			continue
+		}
+		width, widthErr := strconv.Atoi(fields[1])
+		height, heightErr := strconv.Atoi(fields[2])
+		if widthErr == nil && heightErr == nil {
+			sizes[id] = [2]int{width, height}
+		}
+	}
+	return sizes, nil
+}
+
 // Panes returns every managed session's pane pid in a single tmux call,
 // which doubles as a liveness check: a session absent from the map is gone.
 func (d *Driver) Panes() (map[string]int, error) {
--- a/internal/ui/focusscroll.go
+++ b/internal/ui/focusscroll.go
@@ -261,21 +261,39 @@ func (m *Model) requestFocusRegion(sessID string) tea.Cmd {
 }
 
 // focusRegionCmd captures the pane region that sits offset lines above the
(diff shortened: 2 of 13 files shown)


YoanWai:
@coderabbitai review

YoanWai:
@coderabbitai review

PR #376, opened 2026-08-22
fix(ui): let Left leave focus on pi's bare prompt

YoanWai:
## Summary

The focus view's arrow step frees `left` to mean "back to the list" only when the caret sits at the prompt head, and it read that head from the tool's `activity_cutoff` marker. **pi** composes on a bare row between rules and marks it with nothing but its own block cursor, so the cutoff never found a head and `left` was always forwarded into the pane instead of leaving focus.

A new optional per-tool field, `input_prefix`, declares that row directly and replaces the cutoff reuse for this one check. pi's is zero-width on purpose: the live capture strips to a single blank at column zero, so any caret position on the row is the head, and text before the caret is what rules a draft out. The field is backfilled into older configs on load like every other late-arriving one.

## One row of context

A multi-line draft's blank continuation line looks exactly like an empty composer, and a wrapped line looks the same, so the row above decides: it rejects the caret when it carries the same marker with text past it. The rule that bounds the input box is not draft text and does not reject, which is what keeps pi's own divider from refusing its resting state.

Tools that read their marker from `activity_cutoff` are unaffected. The row above either matches the cutoff or carries no marker, and both answer the same as before, so claude, codex, grok and hermes keep exactly the behavior they have today.

## opencode ships no input_prefix

Its live panes park the terminal caret on a fixed blank gutter row while draft text renders *below* it: typing `xy` leaves the caret on the blank row above with only its column tracking. A `┃`-anchored prefix therefore read every gutter row as a prompt head and stole `left` from any draft, not just a multi-line one. One row of context cannot tell those apart, so opencode keeps forwarding `left` exactly as it does today.

## Verification

- Live pi on an isolated tmux socket: at rest the caret is `(0, 42)`, the captured row is `\x1b[7m\x1b[39m \x1b[0m` (strips to one blank), and the row above is pi's `─` rule. Those exact shapes decide `true` at rest, `false` on a draft, `false` at the head of a wrapped line.
- Live opencode on an isolated socket: caret parked at `(5, 44)` on a blank gutter row with four `┃` rows in the block, which is what ruled the opencode default out.
- Fixtures use the shipped default values, so a regression in the default fails a test.
- `gofmt -l .`, `go vet ./...` and `env -u TMUX TMUX_TMPDIR=/tmp/amchk go test ./...` all clean.


YoanWai:
@coderabbitai full review

PR #395, opened 2026-08-23
fix(status): read dialogs, not replayed messages, as waiting

YoanWai:
## What changed

Two pane-reading fixes, both about telling the user's own text from a dialog the tool is drawing right now.

1. Rule matching drops each replayed message from the scope it reads, from its marker row through the rows that wrapped under it. Claude and codex replay a sent message above the composer wearing the same marker the composer carries, so a message that opens with "1." matched the numbered dialog rule.
2. A new `dialog_footer` field names the line only an open dialog draws under that marker. Claude's question dialog puts its selected option on the composer's own row, so the footer is what separates it from a numbered draft. Matching it joins the rows below the marker to what the rules read.

## Why

A session that was mid-turn showed `waiting`. Its pane:

```
✻ Brewed for 6m 49s · 9 messages hidden (/focus to show)

❯ 1. I think we should make a space for marketing & sales right? should we really
  seperate it? because all workers in the company already will have to work on 2 boards

⏺ User message cut off mid-sentence; awaiting clarification
  ⎿  $ source ~/.profile 2>/dev/null

· Razzle-dazzling… (12m 25s · ↓ 30.1k tokens)
```

The prompt's own `❯ 1.` matched the numbered dialog rule, that waiting verdict outranked both the spinner and the `working` hook, and `TypingHold` reported the pane as a dialog, so text sent to the session was held.

The second fix covers the shape the first one's verification turned up. Of the four dialogs captured from a live claude, three indent their selected option and were already read as `waiting`:

```
 Do you want to proceed?
 ❯ 1. Yes
   2. Yes, and always allow access to /tmp/amcap/work from this project
   3. No
```

The question dialog puts it at column 0, where it doubles as the input marker, and only its footer distinguishes it:

```
❯ 1. Spaces
     Fixed-width indent. Renders identical everywhere.
  2. Tabs
  3. Type something.
────
  4. Chat about this

Enter to select · ↑/↓ to navigate · Esc to cancel
```

That pane matched no rule at all, so the session read `idle` and the manager treated it as a composer that would read what it was handed. `dialog_footer` is a new field, and new fields backfill through `mergeTool`, so an existing `config.toml` picks this up.

## How it was verified

Panes captured today from live sessions and from a throwaway claude driven on an isolated tmux socket, then run through the engine with a real user `config.toml` via `config.Load()`:

| pane | before | after |
| --- | --- | --- |
| question dialog (AskUserQuestion) | `idle`, typing allowed | `waiting`, typing held |
| approval dialog | `waiting` | `waiting` |
| plan dialog | `waiting` | `waiting` |
| `/model` picker | `waiting` | `waiting` |
| 18 other live claude, codex, opencode and grok panes | unchanged | unchanged |

Across 22 real panes exactly one verdict moves, and it is the question dialog. The two bug shapes themselves are pinned by unit tests built from those captures, red before each fix and green after.

Tool coverage for the scope change, from live captures rather than reading: codex replays messages under `›` at column 0 the same way claude does under `❯` and gets the same fix; opencode draws one `╹` marker per pane and grok one `│ ❯`, so their scope is untouched. Gemini, hermes and pi declare no `turn_end`, so they never reach that branch.

- [x] `go vet ./...` passes
- [x] `go test -race ./...` passes (25 packages)
- [x] `go build ./...` passes

The last box is honest about scope: the change is in the pane-reading engine, and it was exercised by feeding it real panes from running sessions plus the live user config, not by launching a second manager against the shared store and tmux socket, which would resize every live session's window.

## Visual evidence

The pane text above is the evidence this change acts on: it is what the manager reads, and the table shows what it derives before and after.

## Summary by CodeRabbit

- **New Features**
  - Improved status detection for Claude and Codex dialogs, including question prompts and approval selections.
  - Added support for distinguishing typed drafts from dialog selections.
  - Improved handling of Unicode and wrapped conversation text.

- **Bug Fixes**
  - Prevented sent numbered messages from being incorrectly identified as approval dialogs.
  - Improved detection of waiting and typing states in question dialogs.

- **Documentation**
  - Documented the new per-tool dialog footer configuration option.


YoanWai:
@coderabbitai full review

PR #401, opened 2026-08-25
fix(ui): keep a stale wheel claim from clearing history scroll

YoanWai:
## Problem

Scrolling back through a focused pane's tmux history could snap to the bottom mid-read. The focus watcher pushes captures with the pane's cached mouse flags, and those trail an app that left mouse mode by a debounce. While the stale `paneMouse` claim was still set, the preview handler wiped the scroll offset and applied the live frame, yanking the reader out of the history they were reading.

## Fix

The pane's own history tells the stale claim from a genuine wheel owner: a mouse-tracking agent runs on the alternate screen and keeps no tmux history, so history arriving beside a wheel claim means the cached flag is trailing reality. The leftover-offset wipe now only fires on a pane with no history, and a scrolled-back view holds its frame until fresh flags arrive.

## Tests

- `TestStaleMouseClaimDoesNotClearHistoryScroll`: a capture with a wheel claim beside real history keeps the offset and the held frame.
- `TestWheelFocusWalksTmuxHistory`: the real wheel entry point walks tmux history on a plain pane.
- `TestAppMouseClearsScrollback` still passes: a genuine owner (no history) still returns to the bottom.

## Summary by CodeRabbit

* **Bug Fixes**
  * Preserved scroll position when stale mouse-tracking information is received.
  * Prevented scrolled-back views from unexpectedly resetting to the latest content.
  * Improved wheel scrolling through terminal history in plain panes.
  * Added safeguards to ensure previews remain visible until fresh scrolling state is available.

The guard assumes a genuine wheel owner keeps no tmux history (alternate screen). An app that tracked the mouse on the primary screen would sit outside this rule, but no wheel-driven offset can exist while the mouse flag is set, so no behavior hangs on that shape.

### DIFF
--- a/internal/ui/focusscroll_test.go
+++ b/internal/ui/focusscroll_test.go
@@ -169,6 +169,17 @@ func TestFocusWheelScrollsHistory(t *testing.T) {
 	}
 }
 
+// The wheel entry point, not just scrollFocus, has to walk tmux history
+// on a plain pane: this is the path a real wheel notch takes.
+func TestWheelFocusWalksTmuxHistory(t *testing.T) {
+	m, _ := focusedWithHistory(t, "wheel-walk")
+
+	m.wheelFocus(true, m.pane.box.x+2, m.pane.box.y+1)
+	if m.focusScroll == 0 {
+		t.Fatal("wheel did not walk tmux history")
+	}
+}
+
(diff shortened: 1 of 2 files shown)


YoanWai:
@coderabbitai review

PR #404, opened 2026-08-25
chore(deps): bump clipboard to 0.9.0 and sqlite to 1.57.0

YoanWai:
## What changed

The go-minor-and-patch group bumps land by hand, because one of them is breaking: golang.design/x/clipboard 0.9.0 gives Read a context and an error return. The darwin and linux native image readers now map the new ErrNoData ("clipboard reached, nothing in that format") to this package's ErrNoImage, and let real backend failures propagate as errors instead of collapsing into a silent nil.

modernc.org/sqlite moves 1.56.0 -> 1.57.0 with no call-site changes.

## Why

#399 carried the same two bumps and could not compile: dependabot cannot port call sites across a breaking API change.

A behavior improvement falls out of the new API for free: on 0.8.0 every failure read as nil bytes, so an unavailable pasteboard reported "no image" and darwin skipped its JXA fallback. Now only a genuinely empty image clipboard reports ErrNoImage; an unavailable or failing backend surfaces as an error and the platform fallbacks still run.

## How it was verified

- [x] GOOS=darwin and GOOS=linux both compile; vet clean; gofmt clean
- [x] Live on macOS NSPasteboard: screencapture -c then native read returned a valid 419 KB PNG through the ctx/error API
- [x] Live empty-image clipboard: ErrNoData maps to ErrNoImage
- [x] go test -race ./... fully green

Not exercised here: a live Linux/Wayland read (no such display on this machine). The linux file is byte-for-byte symmetric with darwin and compiles under GOOS=linux.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved clipboard image reading on macOS and Linux.
  * Clipboard read errors are now handled more accurately.
  * Empty or unavailable clipboard content is reported consistently as no image.


YoanWai:
@coderabbitai review

YoanWai:
@coderabbitai review

PR #415, opened 2026-08-27
fix(launch): let the rename directive ride a prompt led by a pasted image

YoanWai:
## What

A launch prompt that opens with an absolute path now carries the rename directive and the coordination note on its first message. Only a real slash command, whose name is a single segment, still opens its own message and takes the notes as a follow-up.

## Why

Pasting an image as the first thing in the quick prompt puts the temp file path at the start of the message. `DirectiveEmbeddable` read any leading `/` as a slash command, so a spawn from that prompt deferred `DeferredRenameDirective` into a second message. That message arrived while the agent was already reading the image, landing in its queue ("Press up to edit queued messages") instead of riding the first prompt.

## Verified

- `TestAssembleCarriesTheDirectiveOverAPastedImagePath` covers both sides: a prompt led by a paste path embeds the directive with no pending inputs, and `/compact` / `/land-pr now` still defer.
- `env -u TMUX TMUX_TMPDIR=/tmp/amtest go test ./internal/launch/` passes; `gofmt -l .` is clean.
- Live spawn from a dev build (`agent-manager-dev spawn --tool claude --prompt "<paste path> describe this image…"`): the pane shows one message, the directive followed by the image path, the session renamed itself to `describe-pasted-image`, and no second message was queued.

## Summary by CodeRabbit

- **Bug Fixes**
  - Prompts beginning with pasted image paths are now processed normally instead of being mistaken for slash commands.
  - Actual slash commands continue to receive their existing special handling.
- **Tests**
  - Added coverage to verify correct handling of pasted image paths and slash commands.

### DIFF
--- a/internal/launch/launch.go
+++ b/internal/launch/launch.go
@@ -46,7 +46,21 @@ func coordinationNote(toolName string, tool config.Tool) string {
 // session's first prompt; otherwise auto-named sessions get the rename
 // directive later as its own message.
 func DirectiveEmbeddable(prompt string) bool {
-	return prompt != "" && !strings.HasPrefix(prompt, "/")
+	return prompt != "" && !opensWithSlashCommand(prompt)
+}
+
+// opensWithSlashCommand separates a prompt the agent reads as a command,
+// which has to open the message, from one that merely starts with an
+// absolute path, as a prompt led by a pasted image does. A command name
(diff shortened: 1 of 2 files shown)


YoanWai:
@coderabbitai full review

YoanWai:
@coderabbitai full review

PR #421, opened 2026-09-01
fix(status): ignore Claude update banner after turns

YoanWai:
## What

- Treat Claude's installed-update banner as UI chrome so a completed turn still resolves to `finished`.
- Migrate the exact legacy built-in chrome pattern while preserving user-customized patterns.
- Cover the stale-working-hook pane captured from the reported session.

## Why

Claude can leave a `working` hook behind when a submitted prompt never receives a matching stop event. The pane fallback normally reconciles that state, but the updater banner appeared beneath the completed-turn marker and made the marker look stale, pinning the session at `working`.

## Verified

- `env -u TMUX TMUX_TMPDIR=/tmp/amst20974008 go test ./...`
- `gofmt -l .`
- `go vet ./...`
- `go build ./...`

## Summary by CodeRabbit

- **Bug Fixes**
  - Updated Claude status detection to recognize the “Update installed · Restart to update” banner.
  - Legacy Claude status settings are upgraded automatically while custom settings remain unchanged.
  - Fixed completed Claude sessions being incorrectly reported as still working after an update.

- **Tests**
  - Added coverage for legacy configuration upgrades, custom status preservation, and update-banner session handling.

### DIFF
--- a/internal/config/config.go
+++ b/internal/config/config.go
@@ -203,6 +203,10 @@ func (c *Config) backfillToolDefaults() error {
 // pattern; one edited by hand keeps what its author wrote.
 const busyLineAgentsOnly = `^[✻✳✶✽✢·✦✧+*] Waiting for \d+ background agents? to finish`
 
+// This exact chrome pattern was written before Claude added its updater
+// banner beneath completed turns. Hand-edited patterns remain untouched.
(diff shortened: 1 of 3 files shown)


YoanWai:
@coderabbitai full review

PR #424, opened 2026-09-01
fix(notify): post iTerm2 notifications through the terminal so a click lands on the session

### LINKED ISSUE #423: macOS notification click opens Script Editor instead of the manager
### What happened

1. Ran the manager on macOS in iTerm2 (`TERM_PROGRAM=tmux`, `TERM=tmux-256color`), so the OSC 777 path is skipped and the `osascript` path in `internal/notify/notify.go:129` delivers the notification.
2. A session flipped to Finished and the desktop notification appeared, titled `agent-manager`.
3. Clicked the notification banner.
4. Expected the manager to come to the front, on the session named in the subtitle. Script Editor opened instead.

The notification is delivered under Script Editor's bundle, so macOS routes the click there. Evidence from the Notification Center store (`~/Library/Group Containers/group.com.apple.usernoted/db2/db`), record 23102, decoded payload:

```
app   = com.apple.ScriptEditor2
titl  = agent-manager
subt  = kids-safety-wearable · codex
body  = ● Finished
soun  = Hero
```

Every recent agent-manager notification in that store carries `app_id` 126, `com.apple.scripteditor2`. A banner posted by `osascript` belongs to the Script Editor bundle, and activating it launches that app.

Two parts to this:

- The click should raise the terminal running the manager, and ideally focus the session named in the subtitle.
- The banner carries the Script Editor name and icon rather than ours.

The Ghostty OSC 777 path is unaffected, since the drawing terminal owns that notification and the click lands on the terminal window.

### agent-manager version

0.35.0

### Operating system

macOS 26.5.1

### tmux version

tmux 3.7b

### Which agent CLI

Not tool specific

### Screenshot or pane output

Reproduces directly from the shell, outside the manager:

```
osascript -e 'display notification "● Finished" with title "agent-manager" subtitle "demo · codex" sound name "Hero"'
```

Click the banner it posts and Script Editor comes forward.

YoanWai:
## What changed

Inside iTerm2 the manager now delivers a session's state change as an OSC 9 escape to the terminal, the same shape the Ghostty path takes with OSC 777. iTerm2 posts the banner itself and reveals the session that sent it when the banner is clicked. `TERM_PROGRAM=iTerm.app` marks iTerm2 at the local shell; inside tmux and on a remote host over SSH the marker is `LC_TERMINAL=iTerm2`, which iTerm2 exports and macOS ships in its default `SendEnv LC_*`. A failed write to the terminal still falls through to the OS path. `docs/usage.md` names the new path.

## Why

On macOS the desktop path runs `osascript`, and macOS attributes every `osascript` banner to Script Editor regardless of the caller: the banner carries Script Editor's name and icon, and clicking it opens Script Editor. Notification Center's own store shows it, record 23102 decoded from `~/Library/Group Containers/group.com.apple.usernoted/db2/db`:

```
app  = com.apple.ScriptEditor2
titl = agent-manager
subt = kids-safety-wearable · codex
body = ● Finished
```

`osascript` reports `com.apple.osascript` as its own bundle id, and `tell application id "com.googlecode.iterm2" to display notification` lands under Script Editor too, so the attribution cannot be moved from the call site. The terminal owning the banner is the path that makes a click land on the manager.

Closes #423

## How it was verified

- `env -u TMUX TMUX_TMPDIR=/tmp/amnt go test ./...`: 26 packages ok. `go test -race ./...`: 25 ok on the first run, with `TestReviveStartsTheAgentAgainInALivePane` in `internal/ui` reporting the quit agent as still running under race load; that test passed 3 of 3 alone and the whole `internal/ui` package passed on a race rerun. It does not touch the notification path.
- New tests pin the sequence for the local, tmux and SSH markers, and the fall-through to `osascript` when the terminal write fails.
- Ran the package against the real terminal: a test binary built from this branch with `TERM_PROGRAM=iTerm.app` wrote its stdout to the manager's tty (`/dev/ttys000`, iTerm2). Captured bytes: `1b5d 393b ... 07`, that is `ESC ] 9 ; ● Finished — probe5 · codex BEL`. iTerm2 handled it within the same second, unified log for process `iTerm2`, subsystem `com.apple.UserNotifications`: `Getting notification settings (async)` then `Got notification settings`.
- On the development Mac iTerm2 stops there: it is not authorized to post notifications (its usernoted registration points at an old AppTranslocation copy and it is absent from `com.apple.ncprefs`), so the banner and the click were exercised through iTerm2's source instead: `screenPostUserNotification` calls `notify:` with the session's window, tab and view indices, and `notificationWasClicked` resolves them and calls `[theSession reveal]`. On a Mac where iTerm2 is allowed to notify, that is the click path.
- WezTerm and kitty keep the OS path; they were not exercised here.

- [x] `go vet ./...` passes
- [x] `go build ./...` passes

## Visual evidence

**Before:**

Notification Center record for an agent-manager banner, `app = com.apple.ScriptEditor2` (decoded above).

**After:**

**Not applicable because:** the change is the escape the manager writes and the app that owns the banner; the bytes and the receiving process are quoted above.

## Summary by CodeRabbit

* **New Features**
  * Added iTerm2 notification support using OSC 9 escape sequences.
  * Notifications now correctly target the active terminal window and workspace, including remote sessions.
* **Bug Fixes**
  * Added fallback to native desktop notifications when terminal notification delivery fails.
* **Documentation**
  * Updated notification guidance with iTerm2 behavior and terminal-specific routing details.


YoanWai:
Closing: this only fixes iTerm2, and the fix has to hold for every terminal. Replacing it with a manager-owned macOS notifier (own bundle, UNUserNotificationCenter) so the banner belongs to agent-manager on any terminal and a click brings the terminal running the manager forward.

PR #439, opened 2026-09-04
feat: copy the full turn text with y, not just the last message_start block

zavatskiy:
`LastMessage`/`LastMessageText` anchor to the newest `message_start` marker, built for a one-line row quote. A reply with several marker-led paragraphs (Claude's bulleted sections, for example) has several `message_start` matches, so anchoring to only the last one drops everything before it.

Adds `Engine.FullTurnText`: same `activity_cutoff`-bounded region, keeps every line instead of anchoring to one marker. Shares a new `toolRules.isStructural` helper with `LastMessage`'s marker search (chrome/busy/turn_end lines dropped the same way in both, extracted so a rule added to one is never missed by the other), additionally drops a `user_echo` line (LastMessage never needed to - its anchor already starts past it) and collapses consecutive blank lines to single paragraph breaks.

Wires it to a new `y` key on the session list: copies the selected session's full reply to the system clipboard without attaching. An explicit key rather than a finish-triggered write, which would silently overwrite whatever the user already had on the clipboard and race several sessions finishing in a row for the same write.

Covered by `TestFullTurnText` (a multi-paragraph reply, checked against `FullTurnText` keeping all of it and `LastMessageText` on the same pane keeping only the last paragraph, the difference this exists for) plus the existing `TestLastMessage` unchanged and passing, confirming the `isStructural` extraction didn't touch its behavior.

Tested locally against live claude/codex sessions, including a session where `pane.mouse` is on and tmux keeps no scrollback, and a long-running session with several bulleted sections in one reply.

## Summary by CodeRabbit

- **New Features**
  - Added the `y` keyboard shortcut to copy the selected session’s latest response to the clipboard.
  - Copies preserve paragraph breaks and exclude interface-only content.
  - Entire responses can be copied, including content beyond the currently visible portion.
  - Added clear feedback for successful copies, empty results, invalid selections, unavailable output, and clipboard errors.
- **Documentation**
  - Documented the new keyboard shortcut in the usage guide and keybinding reference.

### DIFF
--- a/docs/usage.md
+++ b/docs/usage.md
@@ -30,6 +30,7 @@ Tell your agent what you want to review in Agent Manager. Your agent will set up
 | `K` / `J` (or `shift+↑` / `shift+↓`) | Reorder session or group among its visible siblings |
 | `m` | Move a session to a group, a terminal into a session, or a group under another group |
 | `r` | Rename session / edit tool; edit group name and default path |
+| `y` | Copy the selected session's newest reply to the system clipboard, without attaching |
 | `x` | Kill the selected session, or every live session under a group: frees the RAM their agents hold, and the rows stay for `v` |
 | `X` | Kill every live session in view |
 | `v` | Revive a dead session, or every dead session under a group |
--- a/internal/agentsession/claude.go
+++ b/internal/agentsession/claude.go
@@ -23,6 +23,21 @@ func ClaudeTranscriptPath(cwd, sessionID string) (string, error) {
 	return filepath.Join(home, ".claude", "projects", claudeProjectSlug(resolvePath(cwd)), sessionID+".jsonl"), nil
 }
 
+// ClaudeProjectDir is the folder ClaudeTranscriptPath's transcript lives
+// in: every transcript Claude Code has ever written for this working
+// directory, one file per conversation. Claude Code opens a new one on
+// its own context compaction, so a session id captured once can go
+// stale mid-conversation - listing the directory (sorted newest first,
+// say) is how a caller finds the transcript actually being written to
+// now instead of trusting a possibly-stale id.
+func ClaudeProjectDir(cwd string) (string, error) {
+	home, err := os.UserHomeDir()
+	if err != nil {
+		return "", err
+	}
+	return filepath.Join(home, ".claude", "projects", claudeProjectSlug(resolvePath(cwd))), nil
+}
+
 // claudeProjectSlug names the per-directory folder Claude Code files a
 // transcript under: the resolved working directory with every character
 // outside [A-Za-z0-9] turned into a dash.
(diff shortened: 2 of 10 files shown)


zavatskiy:
Update: dropped the on-disk-transcript fallback I'd added (see the revert commit) — caught it live returning content from a genuinely different, hours-old conversation after Claude Code silently rotated to a new transcript file mid-session (looks like context compaction triggers this, invisibly to the tmux pane / agent-manager's own restart handling).

Worth flagging since `claudeTail()` in the poller has the same `AgentSessionID`-trusted-forever assumption for the row-quote fallback — just less visible there since it only ever surfaces one flattened line. Might be worth a mtime check (transcript file's own mtime vs. the pane's last-activity) before trusting a stored id as still-live, for that path too. Didn't want to scope-creep this PR into fixing it, figured it's better flagged than silently inherited.

zavatskiy:
Update: re-added the transcript fallback (dropped in the earlier revert), this time with a safety net instead of trusting AgentSessionID outright — every candidate transcript is only used when its reply provably continues what's already on screen (viewport's own tail found inside the file's reply). When the stored id fails that check but a newer transcript in the same project directory passes, the id self-heals via SetAgentSessionID.

Verified end to end against the same live session that exposed the original bug: stale id correctly rejected, current transcript found by rescan, full multi-paragraph reply recovered. Also caught two smaller issues along the way (both in the commit message) — a composer nudge line leaking into FullTurnText uncaught by any chrome rule, and ClaudeTranscriptTail keeping only the newest assistant entry where a full-turn copy needs all of them since the last user entry (added ClaudeTranscriptFullTurn for that).

YoanWai (changes requested):
Thanks for this, and for the honest commit trail. The revert-and-retry on the transcript fallback is exactly how I'd want it handled. `y` is a feature we need: getting a reply out of a session without attaching is the missing half of driving agents from the list.

A few things need to change before merge.

**Compaction does not rotate the transcript.** Checked two local transcripts with two manual compactions each: every `compact_boundary` is appended to the same `.jsonl` under the same `sessionId`, with hundreds of assistant entries after it. `/clear` stays in-file too. So the stored id does not go stale in-session, and the rescan plus `SetAgentSessionID` self-heal has no case to cover. It does have a failure mode: the anchor is 80 runes of text with no session identity behind it, so two claude sessions in one cwd ending on the same closing sentence rebinds A's id to B's file. From then on `v` resumes B and `f` forks B. Please drop the rescan and the store write. The hours-old transcript you saw was most likely a wrong initial capture in a shared cwd, which is a separate bug and worth an issue.

**`FullTurnText` has no start bound.** `activityRegion` is everything above the composer, and `copyLastOutput` captures 300 lines of history, so on any tool that keeps tmux scrollback `y` copies every prior turn. A two-turn claude pane returns `⏺ a1\n\n⏺ a2`; codex the same shape. It needs to start after the last `user_echo` (or last `turn_end` where a tool has none), and it should also drop tool-call rows, `⎿` result rows, and the continuation rows of a wrapped prompt, which `user_echo` only catches on the first line.

**The fallback anchor rarely matches.** The last 80 runes keep the `⏺ ` marker on a short reply and the soft-wrap `\n  ` indent on a long one at 80 columns; the transcript has neither. So at the default width the fallback never fires and `copyLastOutput` reports success with truncated text. Both sides need the same normalisation (strip the marker, collapse whitespace). The scroll-off gate from 3ba11d8 also needs to come back: a fully visible reply matches its own file and gets replaced with raw markdown the user never saw.

**Smaller:**
- Run capture + write in a `tea.Cmd` like `copySelectionCmd` does; `WriteText` can block 3s inside `Update`.
- `y` on a dead session shows raw tmux stderr; the other keys check `Exists` and show `deadSessionHint`.
- `composerHint` belongs in claude's `chrome_line` (with the old-constant rewrite so existing configs pick it up), otherwise the row quote still shows the nudge while the copy drops it.
- `LastMessageText` has no caller; `ClaudeTranscriptFullTurn` duplicates most of `ClaudeTranscriptTail`.
- `docs/usage.md` and `skills/agent-manager/SKILL.md` list the lifecycle keys and need `y`.

Happy to take the transcript-capture bug myself if you want to keep this PR to the key and the region fix.

YoanWai:
Quick nudge on this, just so you know, agent-manager's built-in MCP already lets any session read any other session and talk to it. The `y` flow here is much faster for grabbing a reply, but if you need more context or coordination between sessions you can just tell your agent: see the agent-manager (I usually type a-m) session of "xxx-xxx".

zavatskiy:
Reworked to your review — the fallback is gone entirely (rescan and `SetAgentSessionID` with it), so nothing rebinds a session's id anymore. Thanks for actually checking the compaction claim; you were right, and I'd built on a wrong diagnosis. Taking you up on the offer to own the transcript-capture bug.

`FullTurnText` now starts after the last `user_echo` (or last `turn_end` where a tool has none), so a two-turn pane no longer copies both. Tool-result rows and a wrapped prompt's continuation rows go with it, and a `turn_end` after content closes the turn.

Three things live use turned up, each now covered by a test:

- **`⎿` is the only glyph safe to treat as a result row.** My first pass also matched `└ ├ │`, which quietly gutted every box-drawing table in a reply — borders kept, rows gone.
- A reply can render as plain prose with no `⏺` marker at all, and was being dropped whole as prompt tail.
- A reply taller than the capture has no prompt in frame to trim after — the region opens mid-reply, so every row of it is content.

The rest is in: capture and write in a `tea.Cmd`, `deadSessionHint` on a dead session, `composerHint` into claude's `chrome_line` with the old-constant rewrite, `LastMessageText` and `ClaudeTranscriptFullTurn` dropped, `y` in `docs/usage.md` and the skill.

Known limitation with the fallback out: on claude the copy is bounded by the visible screen, since tmux keeps no scrollback for an alt-screen tool. Thanks for the MCP pointer — that covers session-to-session nicely. The case it does not cover is getting a reply out to something that is not an agent-manager session at all (pasting into a desktop app, an issue, a doc), where the clipboard is the only channel and a long reply comes out truncated. So the transcript-capture fix would still be worth having for `y`, whenever you get to it.

YoanWai:
Landed in #480 with your commit on main, thank you. Every supported CLI copies a reply now, and a pane that gives no turn boundary says so instead of passing a screenful off as the answer. Closing this as superseded. It ships in the next release.

PR #445, opened 2026-09-04
fix(status): read pi's spinner inside the composer border

### LINKED ISSUE #443: pi status stuck since pi 0.85 moved the spinner into the input border
### What happened

1. Update pi to 0.85.0
2. Spawn a pi session from agent-manager, give it a long task
3. Status says "idle" or "finished" the whole time it works. Never "working".

pi 0.85.0 moved the streaming indicator into the top border of the input box (their changelog, earendil-works/pi#8799). The spinner went from

```
 ⠴ Working...

─────────────────────────────────
```

to

```
── ⠹ Working ────────────────────
```

The [tools.pi] working rule wants the spinner at the start of its own line, then a blank line, then the rule. On 0.85 that shape is gone, so the rule never matches and everything falls back to default_status.

Not the footer this time. I removed the hydra line from the capture and the rule still fails. #414 is fine.

Quick check with the 0.35.0 working rule against my captured pane tail (cut at 50 columns, the rules only need 8 dashes). Runs as-is on Python 3.14:

```python
import re
rule = r"(?ms)^[ \t]*[⠋⠙⠹⠸⠼⠴⠦⠧⠇⠏][ \t]+(?:Working|Running|Retrying|Compacting context|Auto-compacting|Context overflow detected, Auto-compacting|Summarizing branch)\b[^\n]*\n[ \t]*\n─{8,}[ \t]*\n(?:[ \t]*\n)*─{8,}[ \t]*(?:\n[^\n]*){2,5}[ \t]*(?:\n[ \t]*)*\z"
pane = """── ⠹ Working ─────────────────────────────────────

──────────────────────────────────────────────────
~
↑116 ↓26k R1.4M W61k CH96.6% $2.862 (sub) 6.2%/1.
hydra:navigator+simplifier hit 97.1% (last 98.8%)"""
print(re.search(rule, pane))                                # None
old = pane.replace("── ⠹ Working ─────────────────────────────────────",
                   "⠹ Working...\n\n──────────────────────────────────────────────────")
print(re.search(rule, old))                                 # matches. only the spinner placement breaks it
```

Fix idea: match the spinner and its label on their own, without the blank line and rule under them. pi moves this around between releases, and their changelog says custom editors keep the old standalone indicator, so both shapes are out there now.

### agent-manager version

0.35.0

### Operating system

macOS 26.5.2

### tmux version

3.7b

### Which agent CLI

pi

### Agent CLI version

0.85.0 (0.84.2 in #407 still drew the spinner on its own line)

### Relevant config

Stock config.toml, regenerated on 0.35.0. No changes.

### Screenshot or pane output

Working-state pane tail while status said "idle":

```
── ⠹ Working ─────────────────────────────────────

──────────────────────────────────────────────────
~
↑116 ↓26k R1.4M W61k CH96.6% $2.862 (sub) 6.2%/1.
hydra:navigator+simplifier hit 97.1% (last 98.8%)
```

pandysp:
## What changed

The `[tools.pi]` working rule accepts the spinner in the composer's top border (`── ⠹ Working ───`) as well as on a line of its own above it. Both shapes are live: pi 0.85 embeds the indicator by default, custom editors keep the standalone one.

Stored configs carry the old working rule verbatim, so `mergeTool` rewrites it to the current default, same as the #414 footer migration. Both old forms are covered: the pre-#414 two-line footer and the widened-footer form a config written on 0.34 or 0.35 carries. A hand-edited rule stays as written.

## Why

pi 0.85 moved the streaming indicator into the editor border (their changelog, earendil-works/pi#8799). The working rule wanted the spinner at the start of its own line, then a blank line, then the rule. That shape is gone, so the rule never matched and a pi session sat at `finished` through its whole turn.

Closes #443

## How it was verified

- [x] `go vet ./...` passes
- [x] `go test -race ./...` passes, except `internal/ui` `TestReviveStartsTheAgentAgainInALivePane`, which fails the same way on untouched `main` in my environment (claude-hooks live pane, unrelated to this change)
- [x] `go build ./...` passes

Tests: `TestPiPanes` gets the reporter's pane verbatim (border spinner, 3-line footer) deriving `working`, and a historical border-spinner frame followed by a resting editor deriving `finished`. Both fail on the old rule. `TestMigratesStalePiFooterRules` gets the 0.34/0.35 working rule and checks it is rewritten to the current default.

Real sessions: a throwaway program loaded my real `config.toml` (written by 0.35.0, so it carries the old rule) through `config.Load` and `status.NewEngine`, then ran `Match("pi", ...)` on live `tmux capture-pane` output. Two pi 0.85 sessions mid-turn read `working` where the running 0.35.0 manager showed `idle`; the on-disk rule came back migrated. Three resting pi panes read `finished`, not matched.

## Visual evidence

Text captures; the fix is a status string, and I run the manager over mosh, so a screenshot would show the same word.

**Before:** manager 0.35.0 on a pi 0.85 session mid-turn

```
$ agent-manager sessions --json | jq '.[] | select(.id=="c0ff356c") | .status'
"idle"
$ tmux capture-pane -p -t %42 | tail -5
── ⠦ Working ───────────────────────────────────────────────
────────────────────────────────────────────────────────────
~
↑135k ↓15k R4.4M CH99.8% $3.338 (sub) 46.7%/272k (auto)
hydra:navigator hit 98.2% (last 99.2%) $2.8636 (50 obs)
```

**After:** patched engine, same pane, real on-disk config

```
on-disk working rule migrated: true
%42: state=working matched=true
```

**Not applicable because:** see above.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved compatibility with newer pi versions by automatically updating older working-rule configurations.
  * Correctly recognizes activity indicators displayed inside or above the composer border as an active working state.
  * Correctly identifies historical composer-border indicators as finished when the associated turn is complete.
  * Prevents changing spinner activity from being treated as a completed or stuck turn during status detection.

### DIFF
--- a/internal/config/config.go
+++ b/internal/config/config.go
@@ -234,6 +234,13 @@ const (
 	newPiFooterTail = `(?:\n[^\n]*){2,5}[ \t]*(?:\n[ \t]*)*\z`
 )
 
+// pi 0.85 moved the streaming indicator into the composer's top border, and
+// the working rule wanted the spinner on a line of its own above that
+// border. This is the widened-footer form a config written on 0.34 or 0.35
+// carries; it and the two-line form above are rewritten to the current
+// working default.
+const oldPiOwnLineWorkingRule = `(?ms)^[ \t]*[⠋⠙⠹⠸⠼⠴⠦⠧⠇⠏][ \t]+(?:Working|Running|Retrying|Compacting context|Auto-compacting|Context overflow detected, Auto-compacting|Summarizing branch)\b[^\n]*\n[ \t]*\n─{8,}[ \t]*\n(?:[ \t]*\n)*─{8,}[ \t]*` + newPiFooterTail
+
 // mergeTool returns user with any zero-value field filled from def.
 //
 // Shell is deliberately not among them. "terminal" is a plausible name for
@@ -302,8 +309,15 @@ func mergeTool(name string, user, def Tool) Tool {
 	if name == "pi" {
 		for i, rule := range user.Rules {
 			switch rule.Pattern {
-			case oldPiIdleRule, oldPiErrorRule, oldPiRateLimitRule, oldPiQuestionRule, oldPiWorkingRule:
+			case oldPiIdleRule, oldPiErrorRule, oldPiRateLimitRule, oldPiQuestionRule:
 				user.Rules[i].Pattern = strings.Replace(rule.Pattern, oldPiFooterTail, newPiFooterTail, 1)
+			case oldPiWorkingRule, oldPiOwnLineWorkingRule:
+				for _, current := range def.Rules {
+					if current.State == "working" {
+						user.Rules[i].Pattern = current.Pattern
+						break
+					}
+				}
 			}
 		}
 	}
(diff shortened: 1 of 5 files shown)


YoanWai:
@coderabbitai full review

YoanWai:
Merged, thank you @pandysp. The border-spinner rule plus the stored-config rewrite land in the next release, along with the follow-up commit that keeps a pi turn working through long quiet stretches. Nice catch on the 0.85 layout change, and thanks for filing #443 with the exact pane shape, it made the rule easy to verify against a live pi 0.85.1.

PR #437, opened 2026-09-03
fix(ui): align compact panes and restore arrow exit

YoanWai:
## What changed

- Keep tmux cursor coordinates available when an agent hides the real cursor, without drawing that hidden cursor in Agent Manager.
- Keep the cursor cell for every well-formed tmux reply and let only the visibility flag follow `cursor_flag`, so a hidden cursor still carries its coordinates.
- Let Command Code use those coordinates and its painted composer state to make Left leave focus safely at an empty prompt.
- Remove trailing dead rows from compact pane captures and bottom-align the useful rows while keeping selection, caret, and forwarded mouse coordinates mapped to the original pane.

## Why

Command Code paints its own composer caret and hides the terminal cursor. Agent Manager 0.35 correctly stopped drawing hidden cursors, but that also discarded the coordinates used by the Command Code arrow-exit guard, so Left was forwarded and appeared dead.

Command Code 1.39.2 also draws its compact resting surface at the top of a tall tmux pane. Because Agent Manager kept the pane tall to preserve scrollback and painted its trailing blank rows, the composer floated above a large empty band.

Closes #431
Closes #432

## How it was verified

- `gofmt -l .` prints nothing.
- `go vet ./...` passes.
- `go build ./...` passes.
- The arrow-exit test now drives its pane state through `applyPaneState`, the path a live pane takes; it fails on the previous head and passes on this one.
- The affected UI matrix passes on the repository-required isolated tmux socket, covering visible cursors, hidden Command Code cursors, Pi, OpenCode, compact pane alignment, full-screen focus, startup loaders, selection, caret painting, and pane crops.
- Compared the regression against a live Command Code 1.39.2 pane: 156x53 cells, content on rows 2-8, and a hidden tmux cursor reported at column 0.
- The untouched baseline and broad suite exposed unrelated existing tmux timing failures in launch-prompt, autorepeat, revive-picker, and pending-rename tests; targeted reruns confirmed the changed behavior is green.

- [x] `gofmt -l .` prints nothing
- [x] `go vet ./...` passes
- [x] `go test -race ./...` passes on the isolated tmux socket
- [x] `go build ./...` passes
- [ ] Ran the branch against real sessions

## Visual evidence

**Before:** The reproduction and pane geometry are recorded in #432.

**After:** Covered at terminal-cell level by `TestPreviewBottomAlignsCompactPane`; the branch was not swapped into the running manager because doing so would resize active user sessions.

## Summary by CodeRabbit

- **Bug Fixes**
[...]

PR #446, opened 2026-09-04
fix(ui): let Left leave focus when pi hides its cursor

pandysp:
## What changed

Keep tmux's cursor position when an app hides the terminal cursor, without drawing it in agent-manager.

A zero-width `input_prefix` can use that known position. This lets pi tell an empty input row from a draft while keeping the exception limited to tools that explicitly declare a bare input row. Other hidden cursors stay ignored.

This PR is stacked on #437.

## Why

pi hides the terminal cursor and paints its own. tmux still reports the right cell with `cursor_flag=0`.

#437 separates a known cursor position from a visible cursor, but `applyPaneState` still only copies x/y when `cursor_flag=1`. Hidden positions therefore arrive as 0,0. Its focus check also accepts hidden positions only for tools where `ParksItsCaret` is true. That is false for pi because it has no `composer_placeholder`.

On pi 0.85.0, #437 still forwards `←` in every case. It appears to do nothing on an empty input because there is nowhere for the cursor to move. Drafts still receive it.

Keeping the real x/y matters once pi is allowed to use a hidden position. Without it, the check would read 0,0 instead of the input row and could take `←` away from a draft.

## How it was verified

- [x] `go vet ./...` passes
- [x] `go build ./...` passes
- [x] Focused cursor and arrow tests pass
- [x] Full race suite passes with an empty `ZDOTDIR`
- [ ] GitHub Actions test passes
- [x] Ran it against a real pi 0.85.0 session on an isolated tmux socket

The exact full race command with my normal zsh config still fails in `TestReviveStartsTheAgentAgainInALivePane`: syntax highlighting adds ANSI codes to the captured command. The same test fails on untouched #437 at `93acda6`. With an empty `ZDOTDIR`, `go test -race ./... -count=1` passes.

## Visual evidence

Text capture is clearer here. The visible change is which view remains after one key press.

**Empty input:** tmux reported `cursor=0,26 flag=0`. `→` showed `FOCUS`; `←` returned to `View`.

**Draft:** after typing `xy`, tmux reported `cursor=2,26 flag=0`. `←` kept `FOCUS` and moved the cursor to `1,26`.

### DIFF
--- a/internal/ui/focuskeys.go
+++ b/internal/ui/focuskeys.go
@@ -168,7 +168,10 @@ func (m *Model) focusSelected() (tea.Model, tea.Cmd) {
 // the composer is empty and its caret at the head of the prompt; a draft
 // replaces the placeholder and Left belongs to the agent.
 func (m *Model) caretAtInputStart(sessID, tool string) bool {
-	if m.engine == nil || !m.pane.cursor.ok || m.pane.forID != sessID || m.scrolledBack() {
+	if m.engine == nil || m.pane.forID != sessID || m.scrolledBack() {
+		return false
+	}
+	if !m.pane.cursor.ok && !(m.pane.cursor.positionOK && m.engine.ParksItsCaret(tool)) {
 		return false
 	}
 	rows := strings.Split(strings.TrimSuffix(m.preview, "\n"), "\n")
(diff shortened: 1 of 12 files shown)


YoanWai:
Reopened against main. Merging #437 deleted the branch this was stacked on and GitHub closed it with it, sorry.

The `applyPaneState` change is already on main via #437. What is left is the pi part, the zero-width `input_prefix` accepting a hidden but known cursor position. It rebases onto main with one trivial test conflict and the UI tests pass, so I'll land it from here with your commit as is.

Thanks for both PRs @pandysp.

YoanWai:
Landed on main as 5ba8c0c via #453, your commit as author. With #445 in the same release, pi gets both the border spinner and the hidden-cursor arrow exit at once. Thanks @pandysp, closing this one as merged from there.

PR #457, opened 2026-09-06
docs(feed): send the v0.36.0 entry to the key bindings thread

YoanWai:
## What changed

The v0.36.0 feed entry now opens discussion #455, the key bindings thread, and its body names that thread instead of a number.

## Why

One discussion per release. The auto-created announcement (#454) duplicated the release page and was deleted, so the entry points at the thread where the defaults are being decided.

## How it was verified

- TestShippedFeedFileIsRenderableAndRetires passes
- body line 5 is 112 chars

## Summary by CodeRabbit

* **Documentation**
  * Updated the first changelog entry’s wording and link to reference the correct discussion.


YoanWai:
@coderabbitai full review

PR #458, opened 2026-09-06
chore(release): open a release discussion only when there is one to write

YoanWai:
## What changed

`release.discussion_category_name` is gone from `.goreleaser.yaml`, so publishing a release no longer opens an Announcements discussion.

## Why

v0.36.0 ended up with two threads: the auto-created copy of the release body (#454, since deleted) and the key bindings thread (#455) that was actually worth discussing. A release gets a hand-written discussion when it has something to talk about, and none otherwise; the feed entry links wherever that conversation is.

## How it was verified

- `goreleaser check` passes on the edited config

## Summary by CodeRabbit

* **Chores**
  * GitHub releases will no longer be associated with the “Announcements” discussion category.

### DIFF
--- a/.goreleaser.yaml
+++ b/.goreleaser.yaml
@@ -31,7 +31,6 @@ changelog:
   use: github-native
 
 release:
-  discussion_category_name: Announcements
   header: |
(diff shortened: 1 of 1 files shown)


YoanWai:
@coderabbitai full review

PR #465, opened 2026-09-06
fix(ui): defer pending input until a turn ends

### LINKED ISSUE #447: Claude's queued-message block reads as agent output, so the reply cell quotes the manager
### What happened

A session launched with a slash-command prompt (or with no prompt) cannot carry the rename directive inside its first message, so `launch` queues it as a pending input. The poller delivers that pending input as soon as `launchPromptTaken` sees the launch prompt echoed, which is a few seconds in, while the agent is still mid-turn.

Claude Code takes the pasted text as a queued message and pins it above the composer for the rest of the turn:

```
✻ Razzmatazzing… (1m 25s · ↓ 147 tokens)

  ❯ Run this exact shell command once, replacing <name> with a short 2-4 word kebab-case name for the broad feature or theme of this whole session (not
    one subtask of a larger feature): agent-manager rename "<name>". Run rename only this once. Do not rename again later in the conversation unless the
    user explicitly asks you to rename; if they do, pick a broad name from context, not a narrow step. Then continue.

────────────────────────────────────────────────────────────────────────────────────────
❯ Press up to edit queued messages
────────────────────────────────────────────────────────────────────────────────────────
```

So for the whole first response the pane reads as if the session is parked on our rename directive, even though the agent is working normally underneath it. The row keeps its generated name (`claude-84cd`) for the same stretch. Both clear only when the first response ends and the agent finally runs the command.

The manager mirrors it in the row. When a capture lands on a frame where Claude has not drawn the `⏺` bullet, `Engine.LastMessage` finds no anchor and falls back to the last content line of the activity region, which is now the tail of our own queued directive. The reply cell then quotes the manager back to the user:

```
q_10  status=working  reply="user explicitly asks you to rename; if they do, pick a broad name from context, not a narrow step. Then continue."
q_11  status=working  reply="Searching skills ecosystem for changelog ⎿  $ npx -y skills find changelog 2>&1 | head -60 (17s) ..."
q_12  status=working  reply="user explicitly asks you to rename; if they do, pick a broad name from context, not a narrow step. Then continue."
```

The cell flickers between the real tool line and our directive for the length of the turn. The same unanchored fallback also picks up Claude chrome on other frames (`● high · /effort`, `tmux focus-events off · add 'set -g focus-events on' to ~/.tmux.conf and reattach for focus tracking`).

**Steps**

1. Create a session whose prompt opens with a slash command that runs for a while, for example `/find-skills is there a skill for generating changelogs from git history?`.
2. Watch the pane. A few seconds in, the rename directive appears above the composer under `Press up to edit queued messages`.
3. It stays there for the whole first response, and the row keeps the generated name.

**Expected**

The rename directive lands when the agent is between turns, so the first response runs on a clean pane and the row quotes the agent rather than the manager.

**Where it comes from**

- `internal/ui/poller.go:832` `maybeSendPendingInput` gates only on `agentAlive`, an activity region and `launchPromptTaken`. Inbox delivery next to it gates on `inboxDeliverable(derived)` (`idle`, `finished`, `waiting`); pending inputs take no equivalent check, so a directive goes into a pane that is mid-turn.
- `internal/status/status.go:351` `LastMessage` returns the last content line of the region when no `message_start` anchor is on screen, and the queued-message block sits inside the region: Claude indents those rows by two spaces, so `activity_cutoff = "(?m)^❯"` cuts at `❯ Press up to edit queued messages` below them instead of above.

**Two things to fix**

1. Hold a pending input until the session is between turns, the way inbox messages already wait.
2. Teach the claude rules about the queued-message block so its rows never read as agent output, and consider requiring the anchor for tools that declare `message_start` instead of falling back to a bare content line.

### agent-manager version

0.35.0

### Operating system

macOS

### tmux version

tmux 3.7b

### Which agent CLI

Claude Code

### Agent CLI version

Claude Code v2.1.261

### Screenshot or pane output

Pane text and derived row values above were captured live from two sessions on 2026-09-05 by polling `tmux capture-pane` every 2s and running the captures back through `status.Engine`.

atirna:
## What changed

Pending inputs now wait until the captured pane is in a between-turn state before agent-manager sends them. This keeps the deferred rename directive out of an active first response.

## Why

The launch prompt could be visible while the agent was already working, so the next pending input was pasted into a live turn and stayed queued in the agent UI.

Addresses the pending-input timing part of #447. Queued-message parsing is separate.

### Required behavior

A pending input remains queued during an active turn and is delivered once the same session returns to a safe prompt.

### Why this approach

The poller now derives the captured pane state before the existing FIFO delivery path and applies the same between-turn and typing-hold checks used for inbox delivery. It does not add provider-specific status rules or change message quoting.

## How it was verified

- [x] `go vet ./...` passes
- [x] `env -u TMUX TMUX_TMPDIR=/tmp/amtest go test -race ./...` passes
- [x] `go build ./...` passes
- [x] The regression keeps a pending directive queued while working, then delivers it while idle

## Visual evidence

The tmux integration test covers queued input during a busy turn, an unfinished draft, and delivery after the prompt becomes idle.

## Summary by CodeRabbit

* **Bug Fixes**
  * Pending input now waits until the session is between turns before being delivered.
  * Input already claimed for delivery can still be sent when the pane is not otherwise ready.
  * Pending input delivery is paused while the engine detects active typing or an unfinished draft in the pane.
  * Pending input resumes delivery after typed draft text is cleared and the pane is ready.

### DIFF
--- a/internal/ui/poller.go
+++ b/internal/ui/poller.go
@@ -525,7 +525,20 @@ func (p *poller) refreshOnce() tea.Msg {
 			agentAlive := !stat.OK || stat.Procs > 1
 			if pane, err := p.tmux.CapturePane(sess.ID); err == nil {
 				paneLastLines[sess.ID], panePrompts[sess.ID] = p.rowLines(sess, pane)
-				sent, err := p.maybeSendPendingInput(sess, pane, agentAlive)
+				derived, err := p.derivePaneStatus(sess, pane, agentAlive, paneHashes)
+				if err != nil {
+					return errMsg{err}
+				}
+				newStatus = derived
+				// Hold the launch state until the agent first paints its pane,
+				// so a just-created session reads "starting up" rather than
+				// flashing idle before it has booted. A grace cap keeps a tool
+				// that never paints from sticking on starting forever.
+				if sess.Status == status.Starting && !paneBooted(pane) &&
+					time.Since(sess.LaunchTime()) < startingGrace {
+					newStatus = status.Starting
+				}
+				sent, err := p.maybeSendPendingInputWhenReady(sess, pane, newStatus, agentAlive)
 				if err != nil {
 					return errMsg{err}
 				}
@@ -538,11 +551,6 @@ func (p *poller) refreshOnce() tea.Msg {
 					}
 					sessions[i].PendingInputs = sessions[i].PendingInputs[1:]
 				}
-				derived, err := p.derivePaneStatus(sess, pane, agentAlive, paneHashes)
-				if err != nil {
-					return errMsg{err}
-				}
-				newStatus = derived
 				// Launch inputs open the conversation, so they go first; a
 				// message from another agent waits its turn behind them.
 				// A launch input sent this tick leaves pane and derived
@@ -553,14 +561,6 @@ func (p *poller) refreshOnce() tea.Msg {
 					if err := p.maybeDeliverInbox(sess, pane, derived, agentAlive); err != nil {
 						return errMsg{err}
 					}
-				}
-				// Hold the launch state until the agent first paints its pane,
-				// so a just-created session reads "starting up" rather than
-				// flashing idle before it has booted. A grace cap keeps a tool
-				// that never paints from sticking on starting forever.
-				if sess.Status == status.Starting && !paneBooted(pane) &&
-					time.Since(sess.LaunchTime()) < startingGrace {
-					newStatus = status.Starting
 				}
 				// Any real transition re-arms the finished alert.
 				if sess.Acked && newStatus != status.Idle && newStatus != status.Finished {
@@ -831,6 +831,13 @@ func (p *poller) clearRecaptureSeen(sessID string) {
 // A durable claim makes automatic delivery at-most-once: after a process or
 // database failure, an ambiguous input is dropped and surfaced rather than
 // risking the same task or slash command running twice.
+func (p *poller) maybeSendPendingInputWhenReady(sess store.Session, pane, derived string, agentAlive bool) (bool, error) {
+	if !sess.PendingInputClaimed && !inboxDeliverable(derived) {
+		return false, nil
+	}
+	return p.maybeSendPendingInput(sess, pane, agentAlive)
+}
+
 func (p *poller) maybeSendPendingInput(sess store.Session, pane string, agentAlive bool) (bool, error) {
 	if len(sess.PendingInputs) == 0 {
 		return false, nil
@@ -849,7 +856,11 @@ func (p *poller) maybeSendPendingInput(sess store.Session, pane string, agentAli
 	if !agentAlive {
 		return false, nil
 	}
-	region, ready := p.engine.ActivityRegion(sess.Tool, ansi.Strip(pane))
+	clean := ansi.Strip(pane)
+	if p.engine.TypingHold(sess.Tool, clean) != "" {
+		return false, nil
+	}
+	region, ready := p.engine.ActivityRegion(sess.Tool, clean)
 	if !ready {
 		return false, nil
 	}
--- a/internal/ui/poller_test.go
+++ b/internal/ui/poller_test.go
@@ -1684,6 +1684,40 @@ func TestPendingInputWaitsForTheLaunchPrompt(t *testing.T) {
 	}
 }
 
+func TestPendingInputWaitsForBetweenTurn(t *testing.T) {
+	m := buildModel(t)
+	if err := m.spawnSession("ready-tool", "ready-tool-abcd", t.TempDir(), "", "", true, false); err != nil {
+		t.Fatalf("spawn: %v", err)
+	}
+	sess, err := m.store.Get(m.sessionRows()[0].ID)
+	if err != nil {
+		t.Fatal(err)
+	}
+
+	sent, err := m.poller.maybeSendPendingInputWhenReady(sess, "❯ ", status.Working, true)
+	if err != nil {
+		t.Fatalf("send while working: %v", err)
+	}
+	if sent {
+		t.Fatal("pending input was delivered during a live turn")
+	}
+	queued, err := m.store.Get(sess.ID)
+	if err != nil {
+		t.Fatal(err)
+	}
+	if len(queued.PendingInputs) == 0 {
+		t.Fatal("pending input was consumed during a live turn")
+	}
+
+	sent, err = m.poller.maybeSendPendingInputWhenReady(sess, "❯ ", status.Idle, true)
+	if err != nil {
+		t.Fatalf("send while idle: %v", err)
+	}
+	if !sent {
+		t.Fatal("pending input was not delivered between turns")
+	}
+}
+
 // A prompt that never reaches the pane, because it scrolled out or the agent
 // never drew it, must not hold pending input past the grace.
 func TestLaunchPromptTakenGivesUpAfterTheGrace(t *testing.T) {


internal/ui/poller.go
```diff
@@ -849,7 +856,11 @@ func (p *poller) maybeSendPendingInput(sess store.Session, pane string, agentAli
 	if !agentAlive {
 		return false, nil
 	}
-	region, ready := p.engine.ActivityRegion(sess.Tool, ansi.Strip(pane))
+	clean := ansi.Strip(pane)
+	if p.engine.TypingHold(sess.Tool, clean) != "" {
```
YoanWai, line 860: **[bug]** Pending inputs now send at the same rest window as inbox messages (`idle` / `finished` / `waiting`), and `SendText` is `pasteAndEnter`. Inbox holds that window when a person is mid-line (`promptCarriesTypedText`) because Enter would submit their draft mixed with the payload. This path only added `TypingHold`, which reports Working/Waiting from rules or a missing activity region, not a human at the composer. After a slash-command first turn the user often starts typing the follow-up on the newly idle prompt; the next poll can paste the deferred rename into that line. That collision was rare when pending input fired mid-turn. It is now the delivery moment.

**Suggestion:** Call `promptCarriesTypedText` next to `TypingHold` in `maybeSendPendingInput` (after the claimed reconcile, before `ClaimPendingInput`) and return `(false, err)` when the composer already carries text, matching `maybeDeliverInbox`.

internal/ui/poller.go
```diff
@@ -831,6 +831,13 @@ func (p *poller) clearRecaptureSeen(sessID string) {
 // A durable claim makes automatic delivery at-most-once: after a process or
 // database failure, an ambiguous input is dropped and surfaced rather than
 // risking the same task or slash command running twice.
+func (p *poller) maybeSendPendingInputWhenReady(sess store.Session, pane, derived string, agentAlive bool) (bool, error) {
```
YoanWai, line 834: **[suggestion]** The durable-claim comment now sits on `maybeSendPendingInputWhenReady`, which does not claim or consume anything. It belongs on `maybeSendPendingInput`. The new helper's non-obvious bit is the claimed bypass (`!sess.PendingInputClaimed &&`), which exists so a crash still reconciles during a live turn rather than waiting for idle.

**Suggestion:** Move the existing comment back onto `maybeSendPendingInput`. If `WhenReady` needs a line, say why a claimed row skips `inboxDeliverable`.

internal/ui/poller_test.go
```diff
@@ -1684,6 +1684,40 @@ func TestPendingInputWaitsForTheLaunchPrompt(t *testing.T) {
...
+	m := buildModel(t)
+	if err := m.spawnSession("ready-tool", "ready-tool-abcd", t.TempDir(), "", "", true, false); err != nil {
+		t.Fatalf("spawn: %v", err)
+	}
+	sess, err := m.store.Get(m.sessionRows()[0].ID)
+	if err != nil {
+		t.Fatal(err)
+	}
+
+	sent, err := m.poller.maybeSendPendingInputWhenReady(sess, "❯ ", status.Working, true)
```
YoanWai, line 1697: **[suggestion]** `TestPendingInputWaitsForBetweenTurn` calls `maybeSendPendingInputWhenReady` with an injected `status.Working` / `status.Idle` and a fake `"❯ "` pane. It never runs `refreshOnce`, never asks `derivePaneStatus` to classify a working frame, and would stay green if the production call at poller.go:541 still went straight to `maybeSendPendingInput`. The launch-prompt test next to it already drives a real pane through refresh; this regression does not.

**Suggestion:** Keep the helper assertion if useful, and add a refresh-path case: a tool whose working rule stays matched (spinner on screen, cutoff still visible, `launchPromptTaken` already true) must keep the pending input queued, then deliver once the pane is idle.

YoanWai (changes requested):
## Summary

The reorder is the right fix for #447's pending-input half: `refreshOnce` now derives pane status (and applies the starting hold) before FIFO delivery, then `maybeSendPendingInputWhenReady` reuses `inboxDeliverable` so an unclaimed input cannot land on a live turn. The claimed bypass is correct (reconciliation only, never a second paste), and the new `TypingHold` check stops a rest-state paste from answering a dialog. Two gaps remain: delivery at rest still skips the inbox typed-text hold, so `SendText` can submit a half-written composer line; and the regression injects status into the helper instead of locking the `refreshOnce` wiring.

## Evidence for the regression gap

Reverting the production call at `internal/ui/poller.go:541` back to `p.maybeSendPendingInput(sess, pane, agentAlive)` (undoing the fix at its call site) leaves `TestPendingInputWaitsForBetweenTurn` green, and the whole `./internal/ui/` package green as well:

```
$ sed -i '' '541s/.*/\t\t\t\tsent, err := p.maybeSendPendingInput(sess, pane, agentAlive)/' internal/ui/poller.go
$ go test ./internal/ui/ -run TestPendingInputWaitsForBetweenTurn -count=1
ok  	github.com/YoanWai/agent-manager/internal/ui	0.833s
$ go test ./internal/ui/ -count=1
ok  	github.com/YoanWai/agent-manager/internal/ui	145.352s
```

Nothing in the suite holds the wiring this PR exists to change.

## Issue counts by severity

- bugs: 1
- suggestions: 2
- nits: 0

atirna:
pushed ec681f5.
`maybeSendPendingInput` now calls `promptCarriesTypedText` right after `TypingHold`, same shape as `maybeDeliverInbox`:

```go
typing, err := p.promptCarriesTypedText(sess, clean)
if err != nil || typing {
    return false, err
}
```

moved the durable-claim comment back onto `maybeSendPendingInput`, and left one line on `WhenReady` for the claimed bypass since that is its only non-obvious bit.

the regression proved nothing. rewrote it to drive `refreshOnce` and a real pane instead of injecting status into the helper. reverting the call site the way you did:

```
--- FAIL: TestPendingInputWaitsForBetweenTurn/busy_line
    poller_test.go:1723: refresh claimed or delivered pending input during a live turn
```

kept a busy-line case alongside the working-rule one, since `TypingHold` already covers the latter on its own and only the busy-line case exercises the new guard.

full `./internal/ui/` is 1,035 passed against 1,032 on the unmodified head, no failures either side.

YoanWai:
Both points from the last round look right to me now: the typed-text hold is in `maybeSendPendingInput`, and the wiring is locked, reverting line 541 to the old call fails `TestPendingInputWaitsForBetweenTurn/busy_line`.

I pushed one commit on top rather than send you round three for a line.

The rest gate reused `inboxDeliverable` (`idle|finished|waiting`), and `errored` is a resting state too: a pane carrying an error line holds its queue with nothing left to release it, so the rename directive, and the whole launch prompt on a send-mode tool like hermes, waits for a turn that already ended. `pendingDeliverable` adds `errored` to that set, with `TestPendingInputLandsOnAnErroredPane` covering it; swap the helper back to `inboxDeliverable` and it fails.

Verified live as well, real Claude session opened with a slash command: the directive stayed off the pane through a 40s first turn and landed once it ended, and a session resting on an error line took its queue immediately.

Inbox messages have the same stall on an errored pane, which predates this PR. I'll open that separately.

Thanks for sticking with this one.

YoanWai:
@coderabbitai review

YoanWai (approved):
Both gaps from the last round are closed, and the errored rest state is covered by the commit I pushed on top.

Verified on `f1cd74d`: CI green (test, gitleaks, govulncheck), reverting the call site at line 541 fails `TestPendingInputWaitsForBetweenTurn/busy_line`, and swapping `pendingDeliverable` back to `inboxDeliverable` fails `TestPendingInputLandsOnAnErroredPane`. Live run with a real Claude session opened on a slash command: the directive stayed off the pane for the whole first turn and landed when it ended.

PR #453, opened 2026-09-05
fix(ui): let Left leave focus when pi hides its cursor

YoanWai:
## What changed

Andreas Spannagel's (@pandysp) commit from #446, rebased onto main with his authorship kept. A zero-width `input_prefix` now lets a hidden but known cursor position count as the caret cell, so pi's bare composer row can tell an empty input from a draft when pi hides the terminal cursor.

## Why

pi hides the terminal cursor and paints its own. #437 kept tmux's coordinates for hidden cursors but only let tools that park a caret use them, and pi has no composer placeholder, so Left was still forwarded on an empty pi input. #446 was stacked on the #437 branch and closed when that branch was deleted at merge; maintainer edits are off on it, so it lands from this branch.

## How it was verified

- `gofmt -l .` prints nothing
- `go vet ./...` passes
- `internal/ui` passes on the isolated tmux socket, including the pi bare-row and hidden-cursor cases
- Live verification against pi 0.85.0 is recorded in #446

Supersedes #446.

## Summary by CodeRabbit

- **Bug Fixes**
  - Improved keyboard focus detection for composer inputs when the terminal cursor is hidden.
  - Bare composer rows with a known cursor position are now recognized correctly, including empty and typed inputs.
  - Focus handling is more reliable for marked prompts and other interfaces that temporarily hide the cursor.
  - Cursor positioning is now validated more accurately when the cursor is hidden, reducing focus-related inconsistencies.

PR #467, opened 2026-09-07
fix(ui): let Left leave focus while pi is working

pandysp:
## What changed

Left now leaves focus on an empty pi composer while a turn is running, not only once the turn has ended. pi's `activity_cutoff` takes the composer's top border with the spinner drawn inside it (`── ⠹ Working ───`) alongside the plain rule, so the arrow-step head check reads it as the input box's edge rather than a draft line above the caret. A stored `config.toml` carrying the old pi `activity_cutoff` verbatim is rewritten to the current default on load, the way the pi working rule already is; a hand-edited one is kept.

One config field and its migration; no Go logic changes outside `mergeTool`.

## Why

pi 0.85 moved the spinner into the composer's top border. #445 taught the status rules that shape, but `caretRowEndsAPromptHead` steps over the row above the caret only when it matches the tool's `activity_cutoff`, which for pi knew only the plain rule. With pi's zero-width `input_prefix` every row carries "the marker", so the spinner border read as a draft line with text past it, and Left stayed with pi for the length of every turn. #453 could not see it: its cases put a plain rule above the composer, which is what an idle pi draws.

### Required behavior

- Focused on pi mid-turn, empty composer, caret at the prompt head: Left leaves focus.
- Focused on pi mid-turn with a draft typed: Left moves pi's caret; focus stays.
- A tall draft whose border carries the scroll indicator (`─── ↑ 2 more ───`) is not an edge: the row under it is the middle of the draft, and Left stays with pi.
- The stored pi `activity_cutoff` an older release wrote is upgraded; a hand-edited one is untouched.
- Read against a whole pane, the cutoff still ends at the bottom rule: pi's activity region stays anchored at the origin, so no status result moves.
- Other tools are untouched: the row above reaches the cutoff check only when it matches their input marker, and their cutoffs are those markers.

### Why this approach

`activity_cutoff` is the field `MatchesActivityCutoff` already documents as "the rows that bound its input box", and the spinner border is that box's top edge; it is where pi 0.85 put the spinner, and the working rule already reads it as the composer's border. Matching on the spinner glyph inside a rule rather than the label list keeps a custom loader message from turning the border back into draft text.

A first cut let the head check step over any `chrome_line` match instead. opencode's shipped `chrome_line` is `^\s*(┃.*)?$`, every gutter row with the draft text on it, so that read Left as free from the middle of a multi-line opencode draft. `chrome_line` means "not content for the status engine"; it does not mean "not a draft". `TestCaretOnOpencodesMultiLineDraftWithShippedDefaults` pins that on the shipped defaults, where the overlap lives (the existing opencode test builds a minimal engine without a `chrome_line`), and goes red against the chrome-based cut.

## How it was verified

- `gofmt -l .` prints nothing
- `go vet ./...` passes
- `go build ./...` passes
- `TestCaretOnPisComposerRowWhileWorking` on the shipped defaults: red before the change, green after
- `go test ./...` on the isolated socket: everything passes except `TestInboxDeliversToARestingAgentWithItsSenderNamed`, which fails identically on untouched `main` (7c456e2) in this environment, with and without `-race`; under a zsh `SHELL`, `TestReviveStartsTheAgentAgainInALivePane` also fails on untouched `main` (the prompt's syntax highlighting puts colour codes inside the string the assertion looks for) and passes with `SHELL=/bin/bash`
- Ran it against a real pi 0.85 session in an isolated harness (separate `HOME`, `TMUX_TMPDIR`, config), one turn running `sleep 240`, the manager focused with the `FOCUS` badge and `── ⠇ Working ───` on screen, caret at column 0 of the empty composer, swapping binaries on the same turn
  - v0.36.0: Left, `FOCUS` badge still shown, agent caret unchanged
  - this branch: Left, back on the list, agent still working
  - this branch with `ab` typed mid-turn: Left moves the agent caret 2→1, focus stays
- Loaded my own stored `config.toml` (written by 0.36.0, old pi `activity_cutoff` verbatim) through `LoadDir`: the pi `activity_cutoff` comes back as the new default
- Holds for pi; every other tool's Left behaviour is unchanged by construction (see Scope)

## Summary by CodeRabbit

- **Bug Fixes**
  - Updated Pi activity detection to support the newer spinner-in-border display.
  - Existing configurations using the previous activity pattern are automatically updated to the current default.
  - Improved recognition of empty composer rows in Pi and multiline composer input in OpenCode, while preserving draft and continuation text handling.

### DIFF
--- a/internal/config/config.go
+++ b/internal/config/config.go
@@ -253,6 +253,13 @@ const (
 // working default.
 const oldPiOwnLineWorkingRule = `(?ms)^[ \t]*[⠋⠙⠹⠸⠼⠴⠦⠧⠇⠏][ \t]+(?:Working|Running|Retrying|Compacting context|Auto-compacting|Context overflow detected, Auto-compacting|Summarizing branch)\b[^\n]*\n[ \t]*\n─{8,}[ \t]*\n(?:[ \t]*\n)*─{8,}[ \t]*` + newPiFooterTail
 
+// The pi activity cutoff written before pi 0.85 moved the spinner into the
+// composer's top border. It knows only the plain rule as the input box's
+// edge, so the border with a spinner in it read as draft text and Left
+// stayed with the agent for the length of every turn. Hand-edited patterns
+// remain untouched.
+const oldPiActivityCutoff = `(?ms)\A.*^─{8,}[ \t]*$`
+
 // mergeTool returns user with any zero-value field filled from def.
 //
 // Shell is deliberately not among them. "terminal" is a plausible name for
@@ -319,6 +326,9 @@ func mergeTool(name string, user, def Tool) Tool {
 		}
 	}
 	if name == "pi" {
+		if user.ActivityCutoff == oldPiActivityCutoff {
+			user.ActivityCutoff = def.ActivityCutoff
+		}
 		for i, rule := range user.Rules {
 			switch rule.Pattern {
 			case oldPiIdleRule, oldPiErrorRule, oldPiRateLimitRule, oldPiQuestionRule:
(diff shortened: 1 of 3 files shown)


YoanWai:
Hey, thanks for this! Having a Pi user like you contributing regularly is gold for me, so thanks again. I am a bit busy right now, but I'll try to get to this as soon as possible. This flow might be the most important thing in this TUI, so it's important it works in all of our supported tools.

Issue #110, opened 2026-07-30
Add mouse support


YoanWai:
Add mouse interaction support for the agent-manager interface.

tomatotomata:
I’m thinking of taking this on as focused list interaction first: enable Bubble Tea mouse reporting, map clicks to the rendered session/group rows, and use a second click to activate the same action as Enter without changing focused-pane text selection. I’ll add hit-testing tests against the existing row geometry and keep pane mouse ownership intact. Let me know what you think.

YoanWai:
Sure, go for it.

Two things worth knowing first. The selection tradeoff runs the other way from how you framed it: the pane already owns the mouse in focus mode (`internal/ui/focuskeys.go:143`), so nothing changes there. It's the list that has native selection today, deliberately (`main.go:168`, `internal/ui/split.go:87`), and turning reporting on takes it away across the whole rail and detail column. I'd rather that ship as a setting people opt into than a flip everyone gets, so anyone who copies a session name out of the rail keeps doing it. If you think shift-drag covers it well enough, make the case, I'm open.

There also isn't row geometry to hit-test against yet. `m.pane.box` is preview-only (`internal/ui/listview.go:561`). The rail paints a scrolled window of entries that are one or two lines tall depending on density, with `↑ N more` and the ARCHIVED badge pushing everything below them down (`internal/ui/listview.go:117`). So record the index-to-line-range map while painting, the way `pane.box` is recorded, instead of recomputing it in the handler. Recomputed geometry drifts the first time the layout moves.

Two things you do already have: the list-mode mouse handler with divider drag (`internal/ui/split.go:166`), and alt-scroll (`internal/ui/altscroll.go`), which lends the wheel out as arrow keys while reporting is off. Those two are mutually exclusive, so whatever turns reporting on has to turn 1007 off and back.

Second click to activate is the right call, it keeps a double-click timer out of the update loop. No need to claim the issue formally, open the PR when it's ready and I'll review it. One PR, and spell out in the description what you actually ran it against.

YoanWai:
Wheel navigation in the list is disabled ahead of this work (#215). It was never real mouse support: mouse reporting is off outside focus mode, and alternate scroll (DECSET 1007) had the terminal turn each wheel notch into an arrow key that the list handled as cursor movement, so scrolling silently retargeted the next keystroke. 1007 is now off, leaving the wheel inert in the list until this lands. Focus mode is untouched.

Scope for the real thing:

[...]

PR #475, opened 2026-09-08
feat(ui): finish the mouse interaction model (#110)

fruch:
## Summary

Closes the remaining scope from #110. Wheel scrolling only reached the diff (#215, #218); the list swallowed it, same as a click on a session row and a press on the sessions/sidebar divider outside keyboard-armed resize mode.

- A wheel notch in the list moves the session cursor, the same as an arrow key, reusing `moveCursor`'s debounced preview scheduling.
- A click on a session row selects it. `entryLines`/`railLines` record which `m.rows` index each painted line belongs to as they paint (`m.railHits`), the way `m.pane.box` already does for the focused pane, so the click handler never re-derives the rail's layout and can't drift from it, per your comment on the issue. Covered for the split layout, the full screen layout, multi-line comfortable rows, and the "N more" scroll chrome (which correctly picks nothing).
- A press on the divider arms the drag on the spot, whether or not `|` armed resize mode from the keyboard first.
- Settings gets a `mouse` toggle, on by default, for anyone who wants native click-drag text selection over the rail and content column back. It releases mouse reporting everywhere except focus mode, whose own forwarding predates this and stays on regardless.

### Required behavior

- List mode: a wheel notch moves the session cursor up/down; it does nothing while searching or the quick prompt is open.
- List mode: a left click on a session or group row selects it, in the split layout, the full screen layout, and at any list density (compact or multi-line comfortable rows); a click on chrome (search field, filter badges, the "N more" indicators, meters) selects nothing; a click past the rail into the content column selects nothing.
- A left press on the sessions/sidebar divider starts a drag immediately, whether or not resize mode was armed from the keyboard (`|`) first; a press that misses the divider while resize mode is already armed stays a no-op, waiting.
- Settings has a persisted `mouse` on/off toggle (on by default). Off releases mouse reporting for everything this PR touches, giving the terminal back native click-drag text selection over the rail and content column. Focus mode is unaffected either way: its own mouse forwarding predates this setting and always stays on.
- None of the above changes diff-view wheel scrolling, focus-mode mouse handling, or any multiplexer-specific code path (no `termseq`/tmux branching touched).

### Why this approach

- Row hit-testing is recorded at paint time (`m.railHits`, filled by `entryLines`/`railLines`) instead of re-derived in the mouse handler, per your comment on #110 that recomputed geometry drifts the first time the layout moves. This is the same discipline `m.pane.box` already uses for the focused pane.
- Wheel and click both reuse the existing selection path (`moveCursor`/`selectRow`, refactored to share one tail) rather than introducing a second way to move the cursor, so debounced preview scheduling stays in one place.
- The divider-drag arm/disarm reuses `m.split.resizeMode`/`dragging` exactly as the keyboard path does, so a mouse-armed drag and a keyboard-armed drag are indistinguishable to every other code path that reads that state (rendering, `exitResizeMode`, key handling while armed).
- The mouse setting is a single on/off switch releasing capture app-wide (via the existing `syncMouseCapture`, which already did this for the setup dialog) rather than a per-feature toggle, since that is the simplest, most predictable shape and matches how other terminal apps expose "disable mouse."

## What I ran this against

- `go test ./...` (full suite) and `go vet ./...`, both clean. One pre-existing test (`TestPendingInputWaitsForTheLaunchPrompt`) is flaky in my sandbox independent of this change — it fails intermittently on a clean `main` checkout too, unrelated to anything here.
- New tests for every behavior above: wheel-moves-cursor, wheel swallowed while searching/typing in the quick bar, a click selecting a row (split layout, full screen layout, a multi-line comfortable row, a click at the full-layout's phantom divider column, and confirming a click on "N more" chrome is a no-op), a click in the content column not stealing selection, an unarmed divider press starting a drag, a press that misses the divider while resize mode is already armed staying a no-op, and the settings toggle round-tripping through a restart.
- I built this branch and had the person who originally hit #110 run it by hand in the zellij session that showed the bug: wheel scroll, row click, an unarmed divider drag, and the settings toggle releasing/reclaiming native click-drag selection all worked there.
- I did not separately drive it by hand under plain tmux. Nothing in this diff touches multiplexer-specific code (no `termseq`/tmux branching anywhere) — it's plain bubbletea mouse events — and tmux is already exercised throughout the existing suite via the real per-session panes `createSession` spins up.

## Visual evidence

No screenshot or recording attached. A TUI mouse interaction (wheel notches, a click, a drag) doesn't reduce to a single before/after frame the way a static layout change would, and I don't have a way to capture a terminal recording of real mouse events from this environment. The manual verification above (by the person who hit #110, in the zellij session that showed the bug) is the evidence I have; happy to add a recording if that's wanted before merge, or if you can point me at how you usually capture these.

## Test plan

- [x] `go build ./...`
- [x] `go vet ./...`
- [x] `go test ./...`
- [x] Manual: wheel scroll in the session list (zellij)
- [x] Manual: click a session row to select it (zellij)
- [x] Manual: divider drag without pressing `|` first (zellij)
- [x] Manual: settings mouse toggle releases/reclaims native selection (zellij)

## Summary by CodeRabbit

* **New Features**
  * Added mouse controls for selecting sessions, navigating lists with the scroll wheel, and resizing the sidebar by dragging.
  * Added support for selecting hidden sessions through list counters.
  * Added a saved mouse setting to enable or disable app mouse capture, allowing native terminal selection when disabled while preserving focused-pane mouse interactions.
  * Mouse wheel navigation now clamps at list boundaries.

* **Documentation**
  * Expanded mouse-navigation guidance, including selection, resizing, settings, terminal behavior, and focused-pane interactions.


fruch:
@YoanWai I hope this one helps, it's works on my desk ™️

internal/ui/split.go
```diff
@@ -223,22 +206,84 @@ func (m *Model) handleMouse(msg tea.MouseMsg) (tea.Model, tea.Cmd) {
...
+// session row: dragging the seam has to win over the row underneath it.
+// Neither needs resize mode armed from the keyboard first. A press while
+// resize mode is already armed but off the divider is left alone, waiting,
+// exactly as it did before the mouse could arm it too.
+func (m *Model) handleMousePress(msg tea.MouseMsg) (tea.Model, tea.Cmd) {
+	if msg.Button != tea.MouseButtonLeft || m.searching || m.quick.active {
+		return m, nil
+	}
+	y0, y1 := m.bodyYRange()
+	onDivider := m.mode == modeList && msg.Y >= y0 && msg.Y < y1 && m.onDivider(msg.X)
```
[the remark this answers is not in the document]
fruch: Confirmed, real bug: full layout has no divider but `splitWidths()` still returns a ratio-based column, so a click there armed a drag instead of selecting the row. Fixed in the latest push with your exact suggestion (`!m.fullRows()`), plus a regression test (`TestClickAtDividerXInFullLayoutSelectsRow`) that I verified fails without the fix and passes with it.

internal/ui/settings.go
```diff
@@ -188,6 +188,16 @@ func storedHideStats(st *store.Store) bool {
 	return chosen == "on"
 }
 
+// storedMouseDisabled reads the persisted mouse-reporting choice. On is the
+// default; only an explicit "off" gives the rail back to the terminal.
+func storedMouseDisabled(st *store.Store) bool {
+	chosen, err := st.Setting(mouseSetting)
+	if err != nil {
+		return false
```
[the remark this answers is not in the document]
fruch: This matches the existing pattern for every other boolean setting reader in this file (storedArrowStep, storedHideHeader, storedHideStats, storedFullLayout, storedComfortableRows, storedFocusOnEnter, storedNotifications, storedNotifyFinished) - all of them return the default on any Store.Setting error, not just ErrNoRows, without surfacing it. I don't think storedMouseDisabled should be the one function in the file that behaves differently; that'd be its own inconsistency. Happy to fix it, but as a separate pass across all of them rather than scoped to just this PR's new setting - want me to open a follow-up issue for that instead?

YoanWai:
Hey, thanks for this! Mouse in the list has been the leftover #110 hole for a while, and this writeup makes the remaining scope easy to review. I am a bit busy right now, but I will try to get to this as soon as possible.

YoanWai:
Hey, thanks for this! alot of people will probably like this, I am a bit busy right now, but I will try to get to this as soon as possible.

internal/ui/split.go
```diff
@@ -223,22 +206,87 @@ func (m *Model) handleMouse(msg tea.MouseMsg) (tea.Model, tea.Cmd) {
...
+// exactly as it did before the mouse could arm it too.
+func (m *Model) handleMousePress(msg tea.MouseMsg) (tea.Model, tea.Cmd) {
+	if msg.Button != tea.MouseButtonLeft || m.searching || m.quick.active {
+		return m, nil
+	}
+	y0, y1 := m.bodyYRange()
+	// Full layout paints no divider; splitWidths still returns a ratio-based
+	// column, so without this check a click there would arm a drag instead
+	// of landing on the rail row it's actually over.
+	onDivider := m.mode == modeList && !m.fullRows() && msg.Y >= y0 && msg.Y < y1 && m.onDivider(msg.X)
```
YoanWai, line 222: **blocking** &middot; clicking a session row can grab the divider instead

`splitHitSlop` is 1, so the grab target covers `dividerX-1` through `dividerX+1`. `dividerX-1` is the rail's own last painted column, and this check runs before the row does, so that column never makes it to `clickRow`. That didn't matter while the rail wasn't clickable. It does now.

clicking a row grabs the divider

Seam turns into a grip, footer flips to Resize, and the selection never leaves `terminal-dd03`.

| click column | carries row text | what happens |
| --- | --- | --- |
| 30 | yes | selects the row |
| 31 | yes | selects the row |
| 32 (last rail column) | yes | **no selection, divider moves** |
| 33 (seam) | no | divider moves |
| 34 (bleed) | no | divider moves |

That same press commits too, with no motion at all. `setSplitFromX` runs on the press and the release runs `exitResizeMode(true)`, which persists a new ratio and calls `resizeSessions()` across every live session. One click took stored `split_ratio` from unset to `0.2917`.

It also swallows the keyboard. The press sets `resizeMode`, and `handleKey` drops everything except enter, `|`, esc, quit and left/right while that's on. If the release never lands, say you drag out of the window, the list keyboard is gone. I sent the press on its own and then typed `n`, `T`, `j`, `k` and `/`. Nothing. `q` still works, and it quit the manager.

internal/ui/split.go
```diff
@@ -223,22 +206,87 @@ func (m *Model) handleMouse(msg tea.MouseMsg) (tea.Model, tea.Cmd) {
...
+		return m, nil
+	}
+	switch m.mode {
+	case modeList:
+		if m.searching || m.quick.active {
+			return m, nil
+		}
+		switch msg.Button {
+		case tea.MouseButtonWheelUp:
+			return m, m.moveCursor(-1)
```
YoanWai, line 276: **should fix** &middot; the wheel wraps at both ends

`moveCursor` wraps deliberately for the keyboard. One wheel flick is several notches though, so going off the bottom drops you at the top and every key after that is aimed somewhere you didn't intend. Clamping on the wheel path would leave `j`/`k` alone.

```
detail panel names the selection

on the last row       terminal-fba1   terminal
one notch down        root   0 agents
```

internal/ui/split.go
```diff
@@ -223,22 +206,87 @@ func (m *Model) handleMouse(msg tea.MouseMsg) (tea.Model, tea.Cmd) {
...
-// silently retargeting every keystroke that follows (#110).
-func (m *Model) handleMouseWheel(msg tea.MouseMsg) (tea.Model, tea.Cmd) {
-	if m.split.resizeMode || m.mode != modeDiff {
+// handleMousePress resolves a left press against the divider first, then a
+// session row: dragging the seam has to win over the row underneath it.
+// Neither needs resize mode armed from the keyboard first. A press while
+// resize mode is already armed but off the divider is left alone, waiting,
+// exactly as it did before the mouse could arm it too.
+func (m *Model) handleMousePress(msg tea.MouseMsg) (tea.Model, tea.Cmd) {
+	if msg.Button != tea.MouseButtonLeft || m.searching || m.quick.active {
```
YoanWai, line 215: **should fix** &middot; the quick prompt's wheel disagrees with its own footer

`handleQuickKey` sends up and down to `moveCursor` on purpose so the prompt follows the cursor, and the footer advertises it. Both mouse paths bail on `m.quick.active` though, so the wheel and the click do nothing on the one surface built for retargeting.

```
the quick bar's target line

quick bar open       answer terminal-eef9                                   ○ idle  terminal
after wheel down     answer terminal-eef9                                   ○ idle  terminal
after arrow down     answer terminal-8cb8                                   ○ idle  terminal
```

Footer says `↑↓ switch target`.

internal/ui/split.go
```diff
@@ -223,22 +206,87 @@ func (m *Model) handleMouse(msg tea.MouseMsg) (tea.Model, tea.Cmd) {
...
+// handleMouseWheel keeps the wheel inside the app so the outer terminal
+// cannot scroll the manager away: it moves the session cursor in the list,
+// same as an arrow key, and the diff cursor in review.
+func (m *Model) handleMouseWheel(msg tea.MouseMsg) (tea.Model, tea.Cmd) {
+	if m.split.resizeMode {
+		return m, nil
+	}
+	switch m.mode {
+	case modeList:
+		if m.searching || m.quick.active {
```
YoanWai, line 271: **should fix** &middot; nothing moves the selection while search is open

The wheel returns early here, and `handleSearchKey` only takes single rune keys so it throws up and down away. Filter forty sessions down to six and you have to close the field before you can reach any of them. The keyboard side of that isn't new, but the wheel could have fixed it.

```
search open on 'ter', the rail and the selection

before                terminal-eef9   terminal
after wheel + arrow   terminal-eef9   terminal
```

Separately, the new test's comment on this guard is wrong in both directions. It says those two "own the keyboard's up/down too". Search throws them away and the quick bar forwards them.

internal/ui/listview.go
```diff
@@ -346,21 +367,26 @@ func (m *Model) entryLines(rows []treeRow, offset, width, height int) []contentL
...
 	// below only catch an entry that painted taller than entryHeight said.
 	spare := height - len(lines)
 	if start > 0 && spare > 0 {
 		lines = append([]contentLine{{text: subtleStyle.Render(strings.Repeat(" ", railInset) + fmt.Sprintf("↑ %d more", start))}}, lines...)
+		hits = append([]int{-1}, hits...)
 		spare--
 	}
 	if end < len(rows) && spare > 0 {
 		lines = append(lines, contentLine{text: subtleStyle.Render(strings.Repeat(" ", railInset) + fmt.Sprintf("↓ %d more", len(rows)-end))})
+		hits = append(hits, -1)
```
YoanWai, line 383: **consider** &middot; the scroll counters do nothing

Treating `↓ N more` as chrome is the correct hit-test and the test covering it is right. It does leave a gap though. Now that the wheel moves the cursor instead of scrolling the viewport, those counters are the only thing in the rail that still looks clickable, and clicking one gives back a byte-identical frame.

```
short rail, the counter is the last row
█   root         no agents yet   ▐
█ ❯ terminal-6e38  idle · termi… ▐
█ ❯ terminal-eef9  idle · termi… ▐
█ ❯ terminal-8cb8  idle · termi… ▐
█ ❯ terminal-f2f8  idle · termi… ▐
█ ↓ 2 more                       ▐
```

docs/usage.md
```diff
@@ -50,7 +50,7 @@ Tell your agent what you want to review in Agent Manager. Your agent will set up
...
+Navigation is keyboard-driven, but the mouse works too. The manager claims mouse reporting so the wheel stays inside the app and cannot scroll the TUI out of view: a notch moves the session cursor in the list and scrolls the diff, and in a focused session it walks that pane's scrollback, where click-drag also selects pane text and copies it. A click on a session row selects it, and a press on the sessions/sidebar divider drags it without arming resize mode from the keyboard first. Settings has a mouse toggle, on by default, for anyone who wants native click-drag text selection over the rail and content column back instead. Clicking a web link in a focused pane opens it in your browser, in either layout: the mouse claim keeps the terminal's own opener out of reach, so the manager does the opening itself, joining a link the pane wrapped across rows back together first. In a focused agent that tracks the mouse, a click off a link passes straight through to its own clickable UI while a drag still selects and copies; hold `alt` to pass a whole drag through instead, for the agent's own text selection or sliders. See [#110](https://github.com/YoanWai/agent-manager/issues/110).
```
YoanWai, line 53: **should fix** &middot; the mouse toggle reaches further than this line says

This frames it as getting native click-drag selection back over the rail and content column. `syncMouseCapture` releases reporting everywhere except focus mode, review included, so turning it off also kills wheel scrolling in the diff (what #215 and #218 added) and hands the host terminal back the ability to scroll the TUI away, which this same paragraph says the mouse claim prevents. Straight from tmux's record of what the app asked for.

| state | mouse_button_flag | mouse_sgr_flag |
| --- | --- | --- |
| on | 1 | 1 |
| off, list | 0 | 0 |
| off, review | 0 | 0 |
| off, focus | 1 | 1 |

The behaviour's right and the focus carve-out works both ways. It's this sentence that's behind.

internal/ui/model.go
```diff
@@ -177,7 +183,13 @@ type Model struct {
...
 	// uneven height, so every step would re-solve the window and slide the
 	// list under a highlight that should have simply moved down.
-	railTop         int
+	railTop int
+	// railHits maps each line railLines painted this frame to the m.rows
+	// index it belongs to, -1 for chrome (search field, badges, padding,
+	// meters) a click cannot select. Recorded by entryLines/railLines at
+	// paint time, the way m.pane.box is for the focused pane, so a click
+	// handler never has to re-derive the rail's layout and drift from it.
+	railHits        []int
```
YoanWai, line 192: **code shape, optional** &middot; `railHits` ends up with six write sites across three functions in two files

Reset at `listview.go:237`, chrome appends at 244, truncation at 296, the two `entryLines` appends at 349 and 389, and the quick bar pad at 121 up in the view function. Three of them carry a comment whose only job is explaining the lockstep.

`contentLine` is already the per-line carrier, so hanging the row index off the line would let prepend, truncate and pad carry it for free and leave `railLines` the only writer. Would need to be 1-based or paired with a bool, since a bare `contentLine{}` reads as row 0.

internal/ui/listview.go
```diff
@@ -319,14 +335,18 @@ func (m *Model) entryLines(rows []treeRow, offset, width, height int) []contentL
...
 			for _, line := range splitLines(m.renderTreeRow(entry, m.cursor == offset+i, width, offset+i, panelHex())) {
 				lines = append(lines, contentLine{text: line})
+				hits = append(hits, offset+i)
 			}
 		}
 		for _, line := range m.emptyRailLines(width, height-len(lines)) {
 			lines = append(lines, contentLine{text: line})
+			hits = append(hits, -1)
 		}
+		m.railHits = append(m.railHits, hits...)
```
YoanWai, line 349: **code shape, optional** &middot; `entryLines` appends to `railHits` without owning the reset

The reset lives up in `railLines`, the appends live down here, so the lockstep is a convention split across two functions with nothing enforcing it. I measured 34, then 37, then 40 across two direct calls. Five existing test call sites hit this today.

Returning the hits instead of appending to the model would sort this and the six-writers note together.

internal/ui/split.go
```diff
@@ -223,22 +206,87 @@ func (m *Model) handleMouse(msg tea.MouseMsg) (tea.Model, tea.Cmd) {
...
+func (m *Model) clickRow(x, y int) (int, bool) {
+	if !m.fullRows() && x >= m.dividerX() {
+		return 0, false
+	}
+	y0, y1 := m.bodyYRange()
+	if y < y0 || y >= y1 {
+		return 0, false
+	}
+	idx := y - y0
+	if idx < 0 || idx >= len(m.railHits) {
```
YoanWai, line 252: **code shape, optional** &middot; two small ones if you're in the file anyway

`if idx < 0` can't fire. `idx := y - y0` is one line under a guard that already returned for `y < y0`.

And `!m.fullRows()` back at line 222 is just `!m.fullLayout` once `m.mode == modeList` is established. Down here in `clickRow` there's no mode check, so `fullRows()` belongs where it is.

internal/ui/split_test.go
```diff
@@ -466,6 +466,265 @@ func TestPressOffDividerDoesNotDrag(t *testing.T) {
...
+	m := buildModel(t)
+	createSession(t, m, "alpha", t.TempDir(), "")
+	createSession(t, m, "beta", t.TempDir(), "")
+	m.selectSessionRow(t, "beta")
+
+	railWidth, _ := m.splitWidths()
+	m.railLines(railWidth-1, m.listBodyHeight())
+
+	line := -1
+	for i, row := range m.railHits {
```
YoanWai, line 481: **code shape, optional** &middot; the split-layout click tests build the geometry by hand

This and four others work out `railWidth := leftWidth - 1` and `bodyHeight` themselves instead of going through `View()`, which is the thing `clickRow`'s own comment says must never happen. The two full-layout tests already use `View()`.

If `viewListFrame` ever passes different arguments, these five stay green while the real UI picks the wrong row.

YoanWai (changes requested):
Read and driven at `91fc50a`. Build, vet and the full `internal/ui` suite are green.

The core idea here is right. Recording `railHits` while painting, instead of working the geometry out again inside the handler, is what stops this breaking the first time the layout moves. I went looking for a case where the lockstep slips and didn't find one. Groups, three levels of nesting, a scrolled window with both counters showing, an active search, comfortable rows, an empty list, full layout with the quick bar docked. It held in all of them.

You asked how we usually capture terminal evidence. I ran the built binary under a fake `HOME` and its own tmux socket and injected mouse events as raw SGR sequences, so nothing touched a live session or the real store. The frames in the comments below are captures off that run. I can write the recipe up if it's useful.

Findings are on the lines they belong to. One blocking, three worth fixing, one to think about, one docs line. The four tagged **code shape** are optional and none of them are owed.

### Ours already, not this branch

Flagging two so they don't get counted against the PR.

**The `ratioBefore` overwrite predates this.** Arm with `|`, nudge, press the divider, and a later esc restores the nudged ratio instead of the original. `main` sets it the same way and measures identically.

**The settings card already overflows a short terminal.** `cardFlex` only flexes width and `clampFrame` truncates. At 24 rows `main` renders 25 lines and this branch renders 26, so the new row costs one more.

settings card cut off at 24 rows

internal/ui/split.go
fruch: Confirmed, all three parts. Fixed at `68f4701`.

**The grab ate the rail's last column.** `onDivider` reaches right only now — `dividerX` through `dividerX+splitHitSlop`, the seam and the bleed, neither of which carries row text. `dividerX-1` goes back to the row it painted. `TestClickOnRailLastColumnSelectsRow` presses exactly there and asserts both halves: no drag armed, alpha selected.

**The press committed on its own.** It no longer calls `setSplitFromX`, so a press moves nothing; `splitState.moved` is set by motion, and a release with `moved` false is treated as a click — drag cleared, nothing persisted, no `resizeSessions()`. `TestDividerClickWithoutMotionCommitsNothing` presses and releases on the seam and asserts the ratio is untouched and `split_ratio` is still unset in the store.

**It swallowed the keyboard.** A mouse-armed drag now sets `dragging` alone and leaves `resizeMode` — which is the keyboard's gate — off, so `handleKey` never stops seeing keys. The footer and the seam grip read `resizeMode || dragging`, so they look the same as before. If the release is lost, the next key ends the drag where it stands and is then handled normally; `esc` cancels it, exactly as the footer says. `TestKeyEndsADragWhoseReleaseNeverLands` presses the divider, sends `down`, and asserts the drag is gone and the selection moved.

One thing I did fold in while I was in there: the press only takes `ratioBefore` when it is arming fresh (`!resizeMode`). That is the `ratioBefore` overwrite you flagged as ours already — it was one line inside the block I was rewriting anyway, so `|`, nudge, press, esc now restores the original ratio. Happy to pull it back out if you would rather keep it separate.

internal/ui/split.go
fruch: Fixed at `68f4701`. The wheel goes through a new `scrollCursor`, which is `moveCursor` with `min`/`max` instead of the wrap; `j`/`k` and the arrows still call `moveCursor` and still wrap. `TestWheelClampsAtBothEnds` flicks three notches past each end of a three-row list, asserts the cursor held, then calls `moveCursor(-1)` in the same test to pin that the keyboard is still wrapping.

internal/ui/split.go
fruch: Fixed at `68f4701`. Both paths keep the quick bar now: `handleMouseWheel` lost the `m.quick.active` bail entirely, and `handleMousePress` only holds the *divider* while the bar is open — a row click goes through. So the wheel and the click retarget the prompt the same way `↑↓` do, which is what the footer was already promising.

`TestWheelMovesCursorWhileSearchingOrPrompting` and `TestClickSelectsRowWhileSearchingOrPrompting` cover both surfaces (the second opens the bar through `openQuickMode` rather than poking `quick.active`, so the composer is real), and `TestDividerPressBlockedWhileSearchingOrPrompting` pins the divider staying held — the mouse now arms resize under exactly the conditions `enterResizeMode` does.

internal/ui/split.go
fruch: Fixed at `68f4701`, and I took the click with it. The wheel moves the selection while search is open, and so does a click on a row: `railHits` is recorded off the filtered rail the frame actually painted, so clicking one of the six surviving rows lands on the right one. That covers the case you describe from both directions without touching `handleSearchKey`, which still only takes single runes — the keyboard side stays as it was.

You are right about the comment, in both directions. It is gone; the guard it sat on is gone too. The new tests are `TestWheelMovesCursorWhileSearchingOrPrompting` and `TestClickSelectsRowWhileSearchingOrPrompting`, and what they say is what happens.

internal/ui/listview.go
fruch: Took it, at `68f4701` — with one change to the framing you may want to push back on.

A counter now answers for the entry it is hiding: `↑ N more` carries the row just above the window, `↓ N more` the row just below. Click one and the window steps onto that entry, the counter decrements, and you can walk the list with repeated clicks. So the frame is no longer byte-identical, and the only clickable-looking thing in the rail does the thing it looks like.

That does mean the counters carry a row rather than `-1`, which is a different call than the one you said was correct. The way I ended up reading it: the hit test is not "which row painted here" but "which row does a click here select", and under that reading a counter is the one piece of chrome with an obvious answer. `clickRow`'s doc says that now. `TestClickOnMoreIndicatorDoesNotSelect` is replaced by two tests rather than dropped — `TestClickOnMoreCounterSelectsTheRowItHides`, which first asserts the target row is genuinely not painted anywhere in the frame, and `TestClickOnRailChromeDoesNotSelect`, which keeps the search field, badges, padding and meters inert.

If you would rather the counters stayed inert, this is two lines back out and the old test returns as it was.

docs/usage.md
fruch: Fixed at `68f4701`. The sentence now says the toggle releases reporting everywhere except a focused session, and names what that costs: the wheel goes with it, so the diff stops scrolling and the host terminal can scroll the manager out of view again — which is the thing the same paragraph says the claim prevents. Your tmux table is what it is describing.

While in the file I also added that click-to-select now works with the search field and the quick prompt open, and dropped "mouse-driven list navigation" from the README's "not here yet" line, since this is it.

internal/ui/model.go
fruch: Took it, at `68f4701`, your shape exactly. `contentLine` carries a `row int`, one-based so a bare `contentLine{}` reads as chrome rather than as row 0, and `recordRailHits` reads it off the finished line slice once per frame.

The six write sites are two calls: `recordRailHits(railRows)` in `viewListFrame` and in `viewFullListFrame`, each on the last thing the frame assembled. The prepend, the truncation and the quick-bar pad all carry the row for free, so the three comments whose only job was explaining the lockstep are gone with the code they were explaining. `railLines` no longer touches `railHits` at all, and neither does `entryLines`.

internal/ui/listview.go
fruch: Sorted by the same change as the `railHits` note above (`68f4701`). `entryLines` now sets `row` on the lines it paints and returns them; it neither resets nor appends to anything on the model, so calling it twice is just painting twice. `recordRailHits` owns the whole slice and runs once per frame, off the final line list.

The five test call sites you measured 34/37/40 through are also gone — the click tests go through `View()` now, per your other note.

internal/ui/split.go
fruch: Both taken at `68f4701`. `if idx < 0` is gone — you are right, `y < y0` already returned. And line 222 is `!m.fullLayout` now, with `clickRow` keeping `fullRows()` for the reason you give: no mode check down there.

internal/ui/split_test.go
fruch: Taken at `68f4701`. All five go through `View()` now, via a shared `paintedRailLines(t, m, name)` helper that paints the real frame and returns the body lines it put that session on. Nothing in the click tests works out `leftWidth - 1` or a body height for itself any more, so the day `viewListFrame` passes `entryLines` something else, they fail instead of staying green.

`TestClickOnMoreCounterSelectsTheRowItHides` is the one that could not use the helper as-is, since it needs the counter line rather than a session's: it shortens the terminal to force the window, paints through `View()`, and finds the counter in the frame text. It builds no geometry either.

fruch:
All ten addressed, pushed as `68f4701`. The branch is one commit again — the earlier full-layout fix is folded back into the feature commit it belongs to, so there is no `fix: address review` sitting on top.

**Blocking.** All three parts of it. The grab target reaches right only, so `dividerX-1` goes back to the row that painted it; a press no longer calls `setSplitFromX`, and a release with no motion between is a click that commits nothing; and a mouse-armed drag sets `dragging` without `resizeMode`, so the keyboard stays live and a lost release cannot strand the list. Three tests, one per part.

**Should fix.** The wheel clamps (`scrollCursor`), the keyboard still wraps. The quick bar and the search field keep both the wheel and the click — the divider is the only thing still held while they are open, matching `enterResizeMode`. The wrong comment is gone along with the guard it described. The docs sentence now names what the toggle costs, your tmux table being the source.

**Consider.** Taken, but with a change to the framing you may want to argue with: the counters carry the row they are hiding rather than `-1`, so clicking one steps the window onto that entry. That is a different call than "the hit test is correct", and the reasoning is in the thread. Two lines back out if you disagree.

**Code shape.** Took all four. `contentLine` carries a one-based `row`, `recordRailHits` reads it off the finished frame once, and `railLines`/`entryLines` no longer touch the model. The five click tests go through `View()`.

**Ours already.** I folded the `ratioBefore` overwrite in, since it was one line inside the block I was rewriting — the press only takes `ratioBefore` when it is arming fresh, so `|`, nudge, press, esc restores the original now. I left the settings card overflow alone: it is a `cardFlex`/`clampFrame` problem that wants its own change, and this branch still costs it the one extra row. Say the word and I will open an issue for it.

The evidence recipe — fake `HOME`, own tmux socket, SGR sequences injected at the built binary — is worth writing up; I would use it. If you put it somewhere I will start capturing frames that way rather than describing what I saw.

`go build`, `go vet` and `go test ./...` are green. `TestPendingInputWaitsForTheLaunchPrompt` still fails in my sandbox, and I checked it against the merge base `7c456e2` this time: it fails there too, unrelated to anything here.

YoanWai:
Merged, thanks alot! goes out in the next release.

Pushed one commit on top of yours, found it while poking at the drag. If the release never arrives (letting go outside the window) `dragging` stays set, and then the next click on a row hands its own release to the old drag, so the divider jumps to wherever you clicked and saves it. Same thing you fixed for the keyboard in `handleKey`, just the mouse side of it.

Writing the capture setup up in `CONTRIBUTING.md` so it's there next time. Also opened an issue for double click to open a session, take it if you want, otherwise I'll get to it. Thanks again!

YoanWai:
The follow-up is #525.

PR #477, opened 2026-09-09
feat: add Muse Code CLI support

creatorrr:
## What changed

Add Muse Code as a built-in agent CLI (`muse`), including the tool picker, positional startup prompts, installer guidance, status rules, and exact-conversation revive. Capture Muse's session ID from its local session metadata; when an ID is unavailable, revive opens Muse's own session picker. Document the integration and its capabilities.

## Why

Muse users should be able to manage persistent sessions alongside the existing supported CLIs through the same workspace controls.

### Required behavior

- Launch the installed `muse` command with the user's existing defaults and configuration, and pass startup prompts through the shared launch path.
- Recognize idle, working, waiting-for-approval, and error states from Muse's terminal output.
- Match saved conversations by workspace and launch time, exclude subagents and already-claimed conversations, and support snapshot-based recapture after picker selection.
- Keep launch usable when session logs are missing or unrecognized, with the upstream picker as the revive fallback.
- Expose official installation guidance and document the `muse` configuration.

### Why this approach

The integration uses existing tool configuration, launch, status, and revive mechanisms. The only new provider-specific reader extracts the conversation metadata needed for exact-session revive; it honors `XDG_DATA_HOME` and bounds metadata scanning. No models or upstream settings are overridden.

Muse uses the workspace shell subcommands for coordination (`mcp = "none"`): its CLI does not expose an MCP registration interface or a fork command, so neither capability is configured. Existing tools keep their current behavior. Builds cover Linux and macOS on amd64 and arm64; the live Muse session checks ran on Linux amd64, so native interaction on the other three targets remains unverified.

## How it was verified

- [x] `gofmt -l .` prints nothing; `git diff --check` is clean.
- [x] `go vet ./...` passes.
- [ ] Full race suite passes: `env -u TMUX TMUX_TMPDIR=/tmp/amtest go test -race ./...` completed with failures limited to unrelated `internal/feed` and `internal/update` localhost HTTP tests (`connection refused`). The ordinary full suite had the same limitation. Changed-package tests and the UI race checks passed.
- [x] Builds pass with `-buildvcs=false`, required because Go's VCS discovery fails for this `/tmp` worktree. Cross-builds passed for Linux/macOS, amd64/arm64.
- [x] Ran isolated, offline Muse launch and exact-resume probes with a temporary home. The pane displayed `◆ echo: hello`; resume restored the same recorded conversation ID.
- [x] Agent/platform coverage and remaining limitations are named above.

Added regression coverage for default configuration and loading, installation guidance, startup prompt quoting, status classification, tool picker inclusion, and session capture/recapture, including stale/foreign/subagent logs, claimed IDs, ambiguity, missing or malformed logs, and XDG paths.

## Visual evidence

Actual before/after new-session picker captures from the parent build (`055453c`) and this PR build. Each TUI used a fresh temporary configuration and its own tmux socket; no agent sessions were started. These PNGs render the captured terminal text in monochrome, cropped to the unchanged modal bounds; they do not preserve terminal colors.

**Before:** existing picker with Claude selected.

Before: new-session picker

**After:** Muse selected in the same picker.

After: Muse selected in new-session picker

Review follow-up validation: session-capture race tests pass, `gofmt -l .` prints nothing, `go vet ./...` is clean, and `go build -buildvcs=false` succeeds. The local trial binary was rebuilt from the review fix.

## Summary by CodeRabbit

- **New Features**
  - Added Muse Code as a supported tool.
  - Muse Code can be launched, monitored, and resumed by session ID or through its session picker.
  - Added automatic discovery and capture of Muse Code sessions for revival.
  - Included Muse Code in tool selection and quick-spawn ordering.
  - Added default configuration and installer detection support.

- **Documentation**
  - Updated configuration, usage, and supported-tool documentation with Muse Code setup and revival guidance.


YoanWai:
Hey @creatorrr thanks for this! Muse Code is a great addition. Will check this out hopefully by the weekend, since we are currently refactoring how tool definitions are sourced (#478), which relates to any added CLI. No need to adapt to it; I'll probably be the one doing the migration and handling all the supported tools. Thank you again!

internal/config/config.go
```diff
@@ -613,6 +613,27 @@ rules = [
...
+[tools.muse]
+command = "muse"
+session_store = "muse"
+resume_by_id_command = "muse resume {id}"
+resume_picker_command = "muse resume"
+revive_command = "muse resume --last"
+# Muse has no CLI MCP registration interface; use the workspace subcommands.
+mcp = "none"
+default_status = "idle"
+activity_cutoff = "(?m)^⟩"
```
YoanWai, line 625: **Blocking.** The composer marker in Muse 1.1.1 is `❯` (U+276F), not `⟩` (U+27E9). Straight off the pane, `tmux capture-pane` piped through hexdump:

```
e2 9d af 20 | ❯ hello muse       <- the echoed prompt
e2 97 86 20 | ◆ echo: hello muse  <- the reply, so message_start is right
e2 9d af    | ❯                   <- the composer
```

Same bytes on the macOS arm64 and the Linux arm64 builds of 1.1.1.

Because the cutoff never matches, every real pane I captured (idle, draft, mid-turn, settled, picker) comes back from our engine the same way:

```
region=false  hold="working"  lastMessage=""  lastUserEcho=""  draft=""
```

For a Muse session that means: a queued message from another session never arrives (`maybeDeliverInbox` holds while `TypingHold` says working), a queued input never arrives (`maybeSendPendingInput` returns on `!ready`), `y` has nothing to copy, and `derivePaneStatus` never hashes a region, so the session can only ever be working while the spinner row is up, or idle. It never reaches finished, which is the state that raises the turn-done alert.

Fixing the glyph turns all of it on. The same panes then give `region=true hold="" echo="hello muse"`.

The current tests pass because the fixtures spell the marker the same way the config does.

```suggestion
activity_cutoff = "(?m)^❯"
```

internal/config/config.go
```diff
@@ -613,6 +613,27 @@ rules = [
...
+chrome_line = "^\\s*─*\\s*$"
+message_start = "^◆ "
+input_placeholder = "^Type @ to search and insert workspace file paths$"
+user_echo = "^⟩ "
```
YoanWai, line 629: Same glyph here: Muse echoes a submitted prompt on a `❯ ` row (`e2 9d af 20`), so this pattern never matches and `LastUserEcho` stays empty.

```suggestion
user_echo = "^❯ "
```

internal/config/config.go
```diff
@@ -613,6 +613,27 @@ rules = [
...
+command = "muse"
+session_store = "muse"
+resume_by_id_command = "muse resume {id}"
+resume_picker_command = "muse resume"
+revive_command = "muse resume --last"
+# Muse has no CLI MCP registration interface; use the workspace subcommands.
+mcp = "none"
+default_status = "idle"
+activity_cutoff = "(?m)^⟩"
+chrome_line = "^\\s*─*\\s*$"
```
YoanWai, line 626: On macOS the composer sits under a labelled divider, so this bare-rule pattern leaves it as content and it lands in the copy. Real pane through `LastMessage`, with only the glyph fixed:

```
"echo: hello muse hello muse ── Voice input (⌥ + v to start) ────────────────..."
```

Covering the labelled divider gives `"echo: hello muse hello muse"`.

This is why it looked clean on your side: on Linux that row is drawn plain (`────────`), and the label only appears on the macOS build. Same class as the gemini banner we fixed in b9fe29e.

```suggestion
chrome_line = "^\\s*─+(?: .*)?$|^\\s*$"
```

internal/config/config.go
```diff
@@ -613,6 +613,27 @@ rules = [
...
+message_start = "^◆ "
+input_placeholder = "^Type @ to search and insert workspace file paths$"
+user_echo = "^⟩ "
+rules = [
```
YoanWai, line 630: `muse resume` is our revive fallback for a session with no captured id, and it parks on this pane:

```
  Resume a previous session
  Type to search  Filter: [Cwd] All  Sort: [Updated] Created
❯ just now    straw-parallax · Other agent sessions may be running beside you in Agent Manager...
  1m ago      sage-acrux · error: boom
  1 / 3 · 34%  enter resume  esc exit  ctrl+c exit  tab focus sort/filter
```

The highlighted row wears the composer's own marker and no rule here matches it, so once the glyph is fixed the engine reports `status=idle hold="" draft="just now    straw-parallax · ..."`. A queued message would be typed into the picker's filter, and the Enter that ends it would resume whatever was selected. claude and codex stay safe here because their picker rows trip a waiting rule.

The header is stable and matched only this pane out of the six I captured:

```suggestion
rules = [
  { state = "waiting", pattern = "(?m)^\\s*Resume a previous session\\s*$" },
```

internal/config/config.go
```diff
@@ -613,6 +613,27 @@ rules = [
...
+  { state = "waiting", pattern = "(?m)^Do you trust this workspace\\?$" },
+  { state = "waiting", pattern = "(?m)^> \\d+  " },
+  { state = "working", pattern = "(?m)^[◇◈◆] [^\\n]*\\([\\dhms. ]+ · esc to interrupt\\)\\s*$" },
+  { state = "errored", pattern = "(?im)^\\s*error:" },
```
YoanWai, line 634: With no `turn_end`, `matchScope` hands the rules the whole pane, so this fires on any line starting with `error:` anywhere on screen, including text the user has only typed. Real pane, message still sitting unsent in the composer:

```
❯ check this output
  error: boom happened
  > 1  pick me
  done
```

engine: `status=errored matched=true`, and it stays that way while the text is on screen.

claude carries the same pattern but scopes it with a turn_end marker. pi has no marker either and anchors its error rules to the end of the pane (`\z`) for exactly this reason. I could not find an end-of-turn row in Muse's pane, so tail anchoring looks like the way here.

Worth checking the shape first: the one real error I could produce prints with no `error:` prefix at all (`retained session not found: session ... has no saved log`), so this rule may be missing the real thing as well as catching the wrong one.

The `^> \d+  ` rule above is fine as it stands: transcript body rows are indented two spaces, so it only matches the trust prompt's own option rows, which I confirmed on the real dialog.

internal/agentsession/capture.go
```diff
@@ -60,12 +60,15 @@ func resolvePath(p string) string {
...
 // Capture returns the conversation id a tool wrote for a session launched
 // in cwd at or after launchedAt. sessionStore selects the on-disk format
-// ("codex", "opencode", "gemini", "hermes" or "command-code"). claimed holds ids already
+// ("codex", "opencode", "gemini", "hermes", "command-code" or "muse"). claimed holds ids already
 // bound to other sessions, so two sessions started in one directory do not
 // capture the same conversation. It returns ok=false when no confident match
 // exists yet; the caller retries on the next poll.
 func Capture(sessionStore, cwd string, launchedAt time.Time, claimed map[string]bool) (string, bool) {
 	switch sessionStore {
+	case "muse":
```
YoanWai, line 69: Nit, low priority: every other store gets a `captureX` / `snapshotX` / `recaptureX` trio and the switch stays one line per arm. Muse inlines the calls in all three arms instead. Wrapping them keeps this switch uniform with the rest of the file.

YoanWai (changes requested):
Thanks for the patience on this one, and for a careful PR. I took the review the whole way rather than reading it: pulled Muse Code 1.1.1 from the official installer channel (sha256 matched the manifest), ran it offline under `--provider echo` with an isolated HOME, and put the real panes through our status engine. Both the macOS arm64 and the Linux arm64 builds.

What held up against the real CLI:

- The interactive TUI rejects `--session-id` (`invalid TUI options: unexpected argument '--session-id'`), so reading the id back out of the store is the right design here, not extra machinery.
- The store layout, the `subagent/` directories, `runtime.session.metadata`, `workspace_root` and `recorded_at` are exactly what `museMeta` expects. `Capture` against a real store returned the right id, skipped the subagent logs, and took about 0.2ms.
- `muse resume <id>` restores the exact conversation, bare `muse resume` opens the picker, `--last` is real.
- Our launch line works end to end: `muse '<coordination note> ... say hi'` starts the session and submits the prompt.
- The working rule matches the real spinner row (`◈ Thinking (2s · esc to interrupt)`), and both trust rules match the real dialog.
- `muse mcp` is OAuth login only, servers come from `mcpServers` in `settings.json`, so `mcp = "none"` is accurate.
- The installer URL is live and puts `muse` in `~/.local/bin`.

Four fixes in the `[tools.muse]` block, each left inline with the pane it came from:

1. Blocking: the composer marker is `❯` U+276F, not `⟩` U+27E9. The activity region never resolves today, and much of what a session does rides on it.
2. The revive picker reads as a resting prompt, so a queued message would be typed into it.
3. Muse draws no turn-end marker, so rules see the whole pane and `error:` fires on text the user has only typed.
4. `chrome_line` misses the macOS composer divider, which then lands in the `y` copy.

Outside the diff:

- `toolLabel` in `internal/report/report.go` and the tool dropdown in `.github/ISSUE_TEMPLATE/bug_report.yml` both want a Muse Code entry, otherwise a bug filed from a Muse session prefills "A custom tool from my config".
- The branch conflicts with main now in `internal/config/config.go`, `internal/config/config_test.go`, `internal/status/status_test.go`, `docs/configuration.md` and `docs/usage.md`. Main also grew per-tool reply copying since you opened this, so `y` is worth a look on a real Muse pane after the rebase.
- `docs/review/*.png` are review evidence rather than product docs. Attaching them to the PR description keeps main clean.
- Small one, your call: `toolDisplayOrder` puts muse ahead of command-code and hermes, which sort alphabetically after pi today.

Worth knowing: #478 "refactor(config): make Default() the only source of tool definitions" will move where tool definitions live. That migration is mine to do, so nothing to chase here.

Happy to push these fixes onto the branch myself if you would rather I did.

YoanWai:
Hey @creatorrr, checking in on this one. No rush from my side, only that main has moved since and the branch conflicts now, so it gets harder the longer it sits.

The offer still stands: happy to push the four `[tools.muse]` fixes onto the branch myself, plus the `toolLabel` and bug-report dropdown entries, rebase onto main and re-verify `y` on a real Muse pane. Your commit stays on it either way. Say the word and I'll take it from here, or push your own fixes if you would rather keep it yours.

YoanWai:
Hey @creatorrr  last nudge, I will probably take this today unless you want to finish it off

YoanWai:
Thanks a lot for this, @creatorrr! Muse Code support is merged and ships in the next release.

I rebased your two commits onto current `main` with your authorship intact, then added a few follow-ups on top. They tighten the prompt and picker handling for Muse 1.3.0, keep the voice-input divider out of copied replies, and trim the docs. The details are in the PR description.

Really appreciate you taking the time to add it.

internal/agentsession/muse.go
```diff
@@ -0,0 +1,110 @@
...
+		if err != nil {
+			return ""
+		}
+		root = filepath.Join(home, ".local", "share")
+	}
+	return filepath.Join(root, "muse", "sessions")
+}
+
+func captureMuse(root, cwd string, launchedAt time.Time, claimed map[string]bool) (string, bool) {
+	cands, _ := museCandidates(root, cwd, launchedAt.Add(-clockSlack), claimed)
```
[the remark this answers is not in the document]
YoanWai: Fixed after merge in #565 "fix(agentsession): fail a store scan that could not read a conversation". `museMeta` and the codex, command-code and gemini readers now return open, read and close errors. The walk propagates them, snapshot reports the store unavailable, recapture refuses, and capture returns no match on a scan error. An oversized line is still a skipped record.

README.md
```diff
@@ -53,7 +53,7 @@ Not here yet: cost tracking.
 
 ## Supported tools
 
-Status detection supports **Claude Code**, **OpenCode**, **Codex**, **Grok Build**, **Gemini CLI**, **Pi**, **Command Code**, and **Hermes Agent**. Each one's launch, revive, fork, and status rules ship in the binary, so an upgrade brings the current version of all of them (see [Configuration](docs/configuration.md#agent-clis)). A CLI that is not on the list is a [feature request](https://github.com/YoanWai/agent-manager/issues/new/choose).
+Status detection supports **Claude Code**, **OpenCode**, **Codex**, **Grok Build**, **Gemini CLI**, **Pi**, **Command Code**, **Hermes Agent**, and **Muse Code**. Each one's launch, revive, fork, and status rules ship in the binary, so an upgrade brings the current version of all of them (see [Configuration](docs/configuration.md#agent-clis)). A CLI that is not on the list is a [feature request](https://github.com/YoanWai/agent-manager/issues/new/choose).
```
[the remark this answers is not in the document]
YoanWai: Agreed, the wording overclaims. Muse ships no fork command or MCP registration today, and Hermes has no fork either. This moves to the Muse MCP and fork follow-up, which changes these same README and docs lines.

PR #481, opened 2026-09-09
fix(keybind): let a key the file names take it from a default

YoanWai:
A `[keybindings]` table that spends a key on one action collided with another action's default, and the manager refused to open: `config.Load` returned the error, `run()` returned it, `main` printed it and exited, with the user's sessions still running behind it.

Nobody can write that table any other way. An action added after the file was written cannot be moved out of the way in advance, so the day such an action ships a default key is the day everyone already spending it is locked out. `keysBlock` writes every action into `config.toml`, so the people most exposed are the ones who have used the picker.

Found while reviewing #480 (feat(ui): copy a session's newest reply with y), which adds `copy_reply` on `y`:

```
keybindings.list: y is bound to both copy_reply and archive
```

## What changes

What the file names wins. The action that held the key only by default gives it up and is left unbound, where the picker can give it another one.

```toml
[keybindings.list]
kill = "n"     # n kills; new_session is left without a key
```

An action with a second default keeps it, so `review = "ctrl+q"` leaves `detach` on `ctrl+\`. An action the manager needs a key for is still refused when nothing is left: `detach` keeps the way back from a focused session, `settings` keeps the way to the picker. Every other reason a table is refused is unchanged, wording included.

The picker is untouched. It still names what would move and asks, which a file cannot do.

## Two tables now resolve that used to be refused

`kill = "n"` against `new_session`, and `review = "ctrl+q"` against `detach`. The tests that pinned those errors pin the yield instead, plus a new one proving a yield cannot strand the user by emptying `detach`.

## What this replaces

The first version of this PR fell the whole scope back to its defaults and put the reason on the error bar. Review found that loses data: the picker then shows defaults, and `keysBlock` writes every action, so the first save overwrites the user's stored table. Verified on a table holding `fork = "ctrl+f"` and `rename = "ctrl+n"` beside the colliding key - both were gone from `config.toml` after one save. It also dead-ended the repair path, since `SaveKeys` validates both scopes and a broken session table made every list save fail.

Resolving the collision where it arises keeps the user's whole table, leaves only the new action unbound, and needs none of that machinery.

## Still out of scope

A key spec that never becomes a binding (`ctrl+i is tab`, `esc stays as it is`) fails during TOML decode, before any table exists, so it still stops the manager. The same upgrade dynamic applies there the day a release reserves a spelling it used to accept. Closing that means deferring key parsing off `Binding.UnmarshalTOML`, which reshapes what the picker reads; worth its own change.

### DIFF
--- a/docs/configuration.md
+++ b/docs/configuration.md
@@ -47,7 +47,7 @@ A `fork_command` references its source through `{id}` or `{session_file}`, so on
 
 ## Key bindings
 
-Two tables in config.toml name the keys: `[keybindings.session]` for the keys the manager keeps inside a session, and `[keybindings.list]` for the keys of its own list. Each action takes one key or a list of keys, `"none"` turns it off, and an action left out keeps its default. One key serves one action within a table.
+Two tables in config.toml name the keys: `[keybindings.session]` for the keys the manager keeps inside a session, and `[keybindings.list]` for the keys of its own list. Each action takes one key or a list of keys, `"none"` turns it off, and an action left out keeps its default. One key serves one action within a table: a key you name is yours, and an action that only held it by default gives it up and is left without one.
 
 ### Inside a session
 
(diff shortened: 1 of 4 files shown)


YoanWai:
Holding this as a draft: review found the approach loses data. Falling a whole scope back to defaults means the picker then writes defaults over the user's stored table on the first save. Replacing it with a fix in `keybind.build` instead.

Issue #442, opened 2026-09-04
A first-launch mcp add that does not return freezes the manager


YoanWai:
## What happens

The first spawn of a tool that registers MCP through its own CLI runs that registration inside `Update` and waits for it. While the command runs, the manager paints nothing and answers no keys. A registration that returns in the usual moment is invisible; one that does not return holds the whole UI for as long as it takes.

## Where it comes from

`launch.Environment` calls `mcpreg.Apply` (`internal/launch/launch.go:179`), and `Environment` runs on the spawn path inside `Update`. Three commands there wait for a child process:

- `internal/mcpreg/mcpreg.go:268`, `ensureRegisteredOnce`, `cmd.CombinedOutput()`: the one-time `<tool> mcp add --scope user ...` for grok, gemini and command-code.
- `internal/mcpreg/mcpreg.go:239`, the Hermes `hermes mcp add ...`, also `CombinedOutput`.
- `internal/mcpreg/mcpreg.go:205`, `readHermesMCPEntry`, `hermes config get ... --json` through `cmd.Output()`.

None of them carries a context or a deadline, so the wait is unbounded.

It runs once per binary path, gated by the `mcp-<tool>-registered` marker in the hooks directory, so this lands on a fresh install and again after an upgrade moves the binary.

## Two ways it bites

A CLI that sits waiting on the network or on an auth handshake holds `Update` for that whole time.

`CombinedOutput` also waits for the output pipe to close, not only for the process to exit. A registration that leaves a background child holding stdout keeps the manager frozen after the CLI itself has exited.

## How it was found

Driving the real binary under a fake `HOME` and an isolated tmux socket, with a stand-in `grok` on PATH that slept instead of returning. The manager froze: the row stayed at `starting up · 0s ago`, the toast stayed up, and arrow keys did nothing. `ps` showed the wait:

```
/bin/sh /tmp/.../grok mcp add --scope user -e AGENT_MANAGER_SESSION_ID=${AGENT_MANAGER_SESSION_ID} agent-manager -- /tmp/.../agent-manager mcp
```

Killing that wrapper left the UI frozen, because the orphaned `sleep` still held stdout. Killing the `sleep` returned the manager to normal in the same second, which is what pins the cause to the pipe rather than to the process.

## Suggested direction

A deadline is the small version: `exec.CommandContext` with a timeout around each of the three, so a registration that stops answering becomes an error on the status line and the spawn carries on. It keeps the current shape and bounds the freeze.

[...]

PR #489, opened 2026-09-10
fix(mcp): offer the pip line that installs Hermes's MCP SDK

### LINKED ISSUE #487: A Hermes missing its MCP SDK is offered `hermes setup`, which cannot install it
A Hermes whose MCP SDK is missing is refused with a fix that cannot work: the dialog offers `hermes setup`, and `hermes setup` never installs that SDK. Pressing `i` runs it, it exits 0, `hermes` is still on PATH, the launch retries, and the same dialog comes back. There is no way out of the loop from inside the manager.

### What happens

Spawning a Hermes session runs `hermes mcp add`, which answers:

```
Failed to connect: MCP server 'agent-manager' requires the 'mcp' Python SDK, but it is not installed. Run 'hermes setup' to install MCP support, then retry.
```

`internal/mcpreg/mcpreg.go:228` turns that into `ErrHermesMCPUnavailable`, and `internal/ui/launchhint.go` opens the setup-step dialog with `command: "hermes setup"`.

Reproduced on `main` (bc496c1) in an isolated tmux under a fake HOME: the dialog appears, `i` spawns an `install-hermes` terminal running `hermes setup`, the wizard finishes, and the spawn is refused again.

### Why the command cannot work

`hermes setup` is a configuration wizard. Its module (`hermes_cli/setup.py`, Hermes v0.20.0) never installs a package; the only MCP lines it prints are `Add MCP servers: hermes mcp add`. The SDK is an optional dependency of Hermes (`tools/mcp_tool.py:204-217` imports `mcp` inside a try block), and no Homebrew formula vendors it, including the current 2026.8.31: `grep mcp Formula/h/hermes-agent.rb` hits only the bundled `optional-mcps` skill directory.

So on any Homebrew install of Hermes, the manager refuses every Hermes spawn and points at a command that cannot change the outcome.

The retry gate misses it too: after an install `settleInstall` calls `config.CheckInstalled("hermes")`, which passes the moment `hermes` is on PATH, whether or not the SDK arrived.

### What does work

The SDK installs into the interpreter Hermes runs on:

```
/opt/homebrew/Cellar/hermes-agent/2026.8.3_1/libexec/bin/python -m pip install mcp
```

After that, a Hermes session spawned through the manager boots with `agent-manager (stdio) — 25 tool(s)` in its banner.

`hermes --version` names that interpreter's site-packages directory:

```
Install directory: /opt/homebrew/Cellar/hermes-agent/2026.8.3_1/libexec/lib/python3.14/site-packages
```

### Proposed fix

Offer the pip command that installs the SDK into Hermes's own interpreter, resolved from the `Install directory:` line, instead of `hermes setup`. Where the interpreter cannot be resolved, leave the dialog read-only rather than naming a command that does nothing.

YoanWai:
Closes #487.

A Hermes whose optional `mcp` package is missing was refused with `hermes setup`. That command is a configuration wizard: it never installs a package, so pressing `i` ran the wizard, it exited 0, `hermes` was still on PATH, the launch retried, and the same dialog came back.

Which installs hit this: `scripts/install.sh` installs the `[all]` extra, which includes `hermes-agent[mcp]`, so an official install carries the SDK. A `pip install hermes-agent` from PyPI gets the base dependencies only, and the Homebrew formula shipped without the SDK until Homebrew/homebrew-core#302276 (2026-09-05). The message itself is upstream: NousResearch/hermes-agent#109026.

`hermes --version` reports the site-packages of the environment Hermes runs in, which puts its interpreter three levels above:

```
Install directory: /opt/homebrew/Cellar/hermes-agent/2026.8.3_1/libexec/lib/python3.14/site-packages
```

Registration resolves that interpreter where it already runs, off the update path, and the refusal carries the command. The dialog offers it, and `i` runs it in the shell tab the same way a vendor installer runs, so the refused spawn goes ahead on its own once pip is done:

```
Run `'/opt/homebrew/Cellar/hermes-agent/2026.8.3_1/libexec/bin/python3' -m pip install mcp` to add the mcp package to the Python that runs Hermes, then spawn again.
```

The line is offered only where that interpreter answers `pip --version`. pipx and uv build their environments without a pip module, so a Hermes installed that way gets the read-only dialog, which names what has to be installed and leaves the install to the tool that owns the environment.

## Verified

Removed `mcp` from this machine's Homebrew Hermes to reproduce the refusal, then ran this build under a fake HOME on an isolated tmux socket: the dialog showed the pip line, `i` opened the install tab, pip installed the package, the held hermes session launched by itself, and the registration marker was written. Done twice, before and after moving the probe onto the error. `go build ./...`, `go vet` and `go test ./internal/...` pass.

Not exercised: Windows. The manager runs on tmux, and the parser test asserts a POSIX interpreter path, as the existing Hermes tests do for their shell-script fakes.

## Summary by CodeRabbit

- **Bug Fixes**
  - Improved the Hermes MCP setup prompt when the optional MCP SDK is missing.
  - The dialog now provides a `pip install mcp` command for the Python interpreter used by Hermes, replacing the previous `hermes setup` guidance.
  - When the interpreter or pip cannot be identified, the prompt provides generic installation guidance without an automatic install action.


YoanWai:
Thanks, went through all four.

**Fixed: the comment above `fakeHermesVersion`** (6e7728f). It described what the helper does; removed.

**Skipping the Windows interpreter layout.** The manager runs on macOS and Linux, and on Windows through WSL2 (README:56), where paths are Linux paths, so `<prefix>\Lib\site-packages` is not a layout this build parses. A Windows-side Hermes reachable over the WSL interop PATH does not produce a wrong command either: its `Install directory:` is a backslash path, `filepath.Base` on Linux leaves it whole, it fails the `site-packages` check, and the dialog stays read-only. It fails closed, which is the behavior this change is for.

**Skipping the move off the update path, with a caveat.** `hermes --version` costs 0.17s wall here. The call sits on a path that has already blocked `Model.Update` on `exec` for far longer: the same spawn just ran `hermes mcp add` synchronously through `mcpreg.Apply`, which took roughly 20 seconds in the reproduction behind this PR before it returned the error that opens this dialog. So the probe adds no new class of stall, and moving only this one call off the update path would leave the real one in place. The general problem is #442, which tracks a first-launch `mcp add` that never returns freezing the manager, and that is where the whole registration path belongs behind a `tea.Cmd`.

**Skipping the docs and config-comment condition.** Both sentences describe what the dialog does when it has an interpreter to name, which is every supported install. Spelling out the read-only fallback in `docs/usage.md` and in the default config comment would document an edge case at the cost of the sentence that readers need. It is covered by a test, `TestReportLaunchErrorLeavesHermesHintReadOnlyWithoutAnInterpreter`.

YoanWai:
@coderabbitai review

YoanWai:
@coderabbitai full review

YoanWai:
Investigated the update-path finding properly, and you were right. Fixed in e6f1b9d.

The path that makes it real is relaunch: `relaunchInPane` returns a `tea.Cmd` whose comment says the probe and the send run off the update path precisely so a slow pane cannot hold up the UI. `launch.Environment` runs `mcpreg.Apply` inside that goroutine, so `hermes mcp add` never touched `Model.Update`, and the error arrived as `relaunchedMsg`. The probe I added then ran `hermes --version` in `Model.Update` on exactly the path that had been kept clear.

Rather than move the probe into a `tea.Cmd`, the error now carries the command. `ensureHermesRegistered` resolves it where it already runs, and `HermesMCPUnavailableError{PipCommand}` reaches the dialog through `errors.As`, the way `config.MissingToolError` already does in the same function. No exec remains on the update path, and `reportLaunchError` keeps its signature, so the nine callers are untouched.

The version fake in the UI tests is gone with it: those tests now build the error directly.

Verified again end to end: removed `mcp` from this machine's Hermes, ran this build under a fake HOME on an isolated tmux socket, the dialog named the interpreter's pip line, `i` installed it, and the held session launched on its own. Full `go test ./internal/...` passes.

Still skipping the docs condition. Both sentences describe what the dialog does on every install that resolves, and `TestReportLaunchErrorLeavesHermesHintReadOnlyWithoutAnInterpreter` covers the other case.

YoanWai:
Rebased onto `main` after #491. The only conflict was `docs/configuration.md`: that file no longer describes per-tool MCP fields, so the pip-install wording stays in `docs/usage.md`. Stale CodeRabbit `CHANGES_REQUESTED` reviews dismissed.

Issue #482, opened 2026-09-09
OpenCode permission and question dialogs never show as waiting


YoanWai:
### What happened

When OpenCode is blocked on a permission prompt or a question dialog, Agent Manager does not mark the session as waiting. The row stays idle, finished, or working, so the waiting badge, the attention filter, and the waiting notification never fire.

Claude, Codex, and Grok all ship waiting rules for their approval dialogs. OpenCode's built-in rules only cover errors and the working spinner:

```toml
[tools.opencode]
rules = [
  { state = "errored", pattern = "(?i)requires more credits" },
  { state = "errored", pattern = "(?im)^\\s*error\\b" },
  { state = "working", pattern = "(?m)^\\s*▣ +[^·\\n]+· [^·\\n]+$" },
  { state = "working", pattern = "esc interrupt" },
]
```

OpenCode 1.18's TUI draws those dialogs with stable chrome that the rules never look for:

- Permission: `Permission required`, `Allow once`, `Allow always`, `Reject`, `enter confirm` (`packages/tui/src/routes/session/permission.tsx`)
- Question: numbered options, `Type your own answer`, `enter confirm` / `enter submit`, `esc dismiss` (`packages/tui/src/routes/session/question.tsx`)

Plain-text questions that end in `?` above a turn-end duration still resolve to waiting through `turnState`. The modal permission and question overlays do not, so they are invisible in the list.

`matchScope` only pulls dialog chrome that sits below the composer into the match when a waiting rule already matches that footer (`hasWaitingFooter`). OpenCode has no such rule, so even a dialog that replaced the input box cannot flip the row.

Existing `config.toml` files that already list OpenCode rules will not pick up new waiting patterns from backfill (`mergeTool` only copies default rules when the user list is empty). New installs get them; older configs need the patterns added or a targeted merge.

### What I expected

An OpenCode session sitting on a permission prompt or a question dialog should be `waiting`, the same way Claude's "Enter to confirm" and Codex's numbered approval dialogs already are.

### Steps

1. Start an OpenCode session with permissions set to ask.
2. Give it a task that needs a tool approval (a shell command or an edit).
3. When OpenCode shows `Permission required` / `Allow once`, look at the Agent Manager row.

Same for an agent `question` dialog (`Type your own answer`, numbered choices).

### Which agent CLI

OpenCode

### agent-manager version

0.36.0

### Operating system

macOS

### tmux version

tmux 3.7b

### Which agent CLI

Grok Build

[...]

PR #510, opened 2026-09-15
fix(status): read an opencode turn that died on a provider error as errored

### LINKED ISSUE #509: OpenCode turn that dies on a provider error flaps working/finished every 20s
### What happened

When an OpenCode turn ends on a provider error (a rejected API key, or "This model is currently experiencing high demand" once the retries run out), the session row never settles. Measured on main (`8aa2885`, same rules as 0.37.0) with opencode 1.18.31 through the real binary: the row reads working for about 20s, finished for one poll, working again, and repeats for as long as the pane stays like that. Every flip to finished fires the finished notification. errored never shows.

Quick-prompt spawn with a broken Google key, one line per status change:

```
t=6   working
t=32  finished
t=34  working
t=52  finished
t=54  working
t=72  finished
t=74  working
```

### Why

OpenCode draws the turn header `▣  Build · <model>` when a turn starts and appends ` · <duration>` only once the message is final (`packages/tui/src/routes/session/index.tsx`, `final()`). A provider error leaves the bare header behind, under an error box drawn in the same `┃` gutter as a user message, so the plain-text pane looks exactly like a turn that just started. The bare header matches the built-in working rule.

#393 added the stuck grace for this: a rule-matched working pane that stays frozen for 15s settles through `TurnEndedState`. That gives finished, and on the next poll the rule matches working again with nothing on screen changed, so the grace starts over.

The one thing on screen that says whether a turn runs is the footer under the composer. While a turn runs it carries the knight-rider spinner, `■■■⬝⬝⬝⬝⬝  esc interrupt`, and keeps it through provider retries (`⬝⬝⬝■■■ This model is currently experiencing high demand… [retrying in 22s attempt #3`). At rest it shows the working directory and the token count. The rules never look at it.

### What I expected

A turn that dies on a provider error reads errored on the next poll and stays there until the next turn starts.

### Steps

1. Break the Google key in `~/.local/share/opencode/auth.json` (any provider works the same).
2. `space` on a group row, type any prompt, `tab` to opencode, `enter`.
3. Watch the row cycle working, finished, working.

### agent-manager version

main `8aa2885` (0.37.0 ships the same opencode rules)

### Operating system

macOS

### tmux version

tmux 3.7b

### Which agent CLI

OpenCode

### Agent CLI version

opencode 1.18.31

YoanWai:
## What

- New tool field `busy_footer`: the row a CLI paints under its activity cutoff only while a turn runs. A working rule that matches while that footer is back in its resting form belongs to a turn that died before its end marker, and reads `errored`.
- opencode sets it to its knight-rider spinner (`■■■⬝⬝⬝⬝⬝`), which the footer keeps through provider retries.
- The poller's hung-turn fixtures carry the spinning footer, so the 15s stuck grace still covers a pane frozen mid-turn.

## Why

Closes #509. An opencode turn that dies on a provider error leaves its bare `▣  Build · <model>` header behind, which matches the working rule; the stuck grace settles it finished after 15s and the next poll matches working again, so the row flaps working/finished every 20s with a notification each cycle. The footer is the one thing on screen that says whether a turn runs.

## Verified

- `TestDefaultRulesRealPanes` gains five real opencode 1.18.31 captures: provider error → `errored`, turn cut short during a retry → `errored`, slow model → `working`, retry banner → `working`, esc interrupt (`· interrupted`) → unmatched, settled by the quiet path as before.
- Real binary under a fake HOME on an isolated tmux socket, opencode 1.18.31, quick-prompt spawns:
  - broken Google key: `starting`, `working`, `errored` within 4s of the error, stable for the next 40s. Same spawn on main: `working` 26s, `finished`, `working` 2s later, repeating every 20s for two minutes.
  - free OpenCode Zen model with shell allowed: `working`, `finished` at the `· 29.6s` header.
  - free model with shell asked: `working` through the permission dialog, unchanged (#482).
- Unexercised: the turn shape after a rejected permission (the model kept answering in every run, so the header gained its duration), and Linux, where nothing in the footer regex differs.
- `gofmt -l .` prints nothing, `go vet ./...` is clean, full `go test ./...` green on an isolated socket.

## Summary by CodeRabbit

* **Improvements**
  * Improved activity status detection for opencode sessions.
  * Turns that stop unexpectedly during authentication failures or provider retries are now identified as errored.
  * Slow model responses and active provider retries continue to display as working.
  * Interrupted turns are correctly shown as idle.
  * Stuck activity indicators are handled more accurately.

* **Documentation**
  * Expanded usage guidance with details about detecting stalled or errored turns.


YoanWai:
@coderabbitai full review

YoanWai:
@coderabbitai full review

PR #519, opened 2026-09-16
feat(config): add Antigravity CLI (agy) support

### LINKED ISSUE #516: Add Antigravity CLI (agy) support
### What are you trying to do

Google's Antigravity CLI (agy), the successor to Gemini CLI, isn't in the built-in tool list. I drove a live session and verified the following against the real installed CLI (v1.2.0) and its actual terminal output, not docs:

CLI surface (`agy --help`):
- No positional prompt argument. Prompts only via `-p`/`--print`/`--prompt` (one-shot), `-i`/`--prompt-interactive` (initial prompt, stays interactive), or stdin.
- `--conversation <id>` resumes a specific conversation by id; `-c`/`--continue` resumes the most recent.
- No `--session-id` flag, so an id can't be minted at launch (unlike Gemini CLI).
- `/resume` inside the TUI opens a real saved-conversation picker (workspace-grouped, searchable).

UI chrome (captured live via tmux):
- Box-drawing chrome nearly identical to Gemini CLI's.
- Idle composer is a bare `>` (no trailing space when empty, unlike Gemini's `> `).
- Busy footer reads "esc to cancel" (not Gemini's "esc to interrupt").
- Completed turn opens with `▸ Thought for Ns, N tokens` then a bolded title then indented response.
- Tool-permission dialog reads "Run this command?" with a numbered `1. Yes ... 4. No, cancel` list; its footer is also "esc to cancel", so it needs to rank above the generic busy-state match or it reads as merely working instead of waiting on the user.

I built and verified a complete custom tool block against this (back when [tools.*] was still read), fixing three real bugs along the way: prompt embedding (needs a prompt_flag, no positional arg), an activity_cutoff regex that silently deadlocked message delivery forever on an empty composer, and the waiting-dialog/busy-footer ambiguity above. Happy to hand over the exact working TOML block (launch/resume/revive commands + status regexes) if useful as a starting point for a native implementation.

### Anything else

agent-manager 0.36.0 on macOS, reported from a Claude Code session.

mateuszgachowski:
## What changed

Adds a built-in `[tools.antigravity]` block for Google's Antigravity CLI (`agy`), the successor to Gemini CLI: launch (`prompt_flag`), resume-picker, and revive commands, plus status-detection rules (`activity_cutoff`, `chrome_line`, `message_start`, `turn_end`, `user_echo`, and the waiting/working `rules`).

## Why

Closes #516

### Required behavior

Every field was verified live against agy v1.2.0 by driving a real session and reading `tmux capture-pane`, not from docs (details in the code comments and in #516). Two things worth a reviewer's attention:

- `activity_cutoff` is `[>!*]\s?` (optional trailing space), not gemini's required-space form: agy's empty composer is a bare `>` with no trailing space, and a required space there leaves the tool permanently unresolved.
- The `rules` list ranks the "Run this command?" tool-permission dialog above the generic busy-footer match, because both end in "esc to cancel".

### Why this approach

Followed the gemini block as the closest template, per the existing pattern in this file.

**Known gap, not attempted here:** agy has no `--session-id` flag and nothing here captures the conversation id it mints on its own, so restart cannot yet recover a *specific* prior conversation (falls back to the picker or "resume most recent" instead). This is flagged inline in the config and is what `TestRestartLaunchIsAFreshStartForEveryShippedTool` catches (see below). The natural fix is a `session_store` capturer like gemini's, reading agy's own `conversation_summaries.db` (`~/.gemini/antigravity-cli/conversation_summaries.db`) for the newest row after launch - happy to take that on as a follow-up if useful, but wanted to land the verified status-detection piece first rather than block on it.

## How it was verified

- [x] `go vet ./...` passes
- [x] `go test -race ./internal/config/... ./internal/status/...` passes
- [ ] `go test -race ./...` - see note
- [x] `go build ./...` passes
- [x] Ran it against real sessions (drove a live agy session for hours across launch, resume, revive, and the permission dialog while building this)
- [ ] Holds for every supported agent CLI and platform - antigravity itself is untested on Linux/WSL; only macOS

`internal/ui` isn't clean under `-race ./...` on this machine, but confirmed unrelated to this change: the same ~34 tests fail identically on unmodified upstream `main` here too (this box runs a live agent-manager with many real tmux panes, which starves the suite's own tmux spin-ups - reproduced by stashing this diff and rerunning). The one failure that *is* about this change is `TestRestartLaunchIsAFreshStartForEveryShippedTool`, exactly the known gap above.

## Visual evidence

**Not applicable because:** this is a status-detection/launch config change with no UI surface of its own; verification was driving real agy sessions through agent-manager rather than screenshots.


YoanWai:
Hey, thanks for this. The block reads well and the live-verified notes are exactly what I was hoping for.

PR is still marked as a draft, when you're ready flip it and let coderabbit review it also.

On the restart gap, I'm fine landing status detection first and taking the `session_store` capturer as a follow-up. `TestRestartLaunchIsAFreshStartForEveryShippedTool` is the only red check, so let's agree on how to handle it before merge, either here or in the follow up.

YoanWai (changes requested):
Thanks for this, and for the live notes. I installed agy 1.2.9 (the build the installer ships today) in a sandbox and ran the block against real panes, feeding the captured frames through `status.Engine` on this branch rebased onto main. A few things don't hold on 1.2.9.

**1. `turn_end` switches off the working rule**

With `turn_end` set, `matchScope` limits the rules to the rows above the composer. `esc to cancel` is the footer below it, so the working rule never matches. Every busy frame I captured (`Generating...`, `Working...`, `Running command...`) comes back `Match = idle` (unmatched) with `TypingHold = ""`. The row still turns working in the UI only because the poller treats a changing region as activity.

The `▸ Thought for 2s, 435 tokens` line isn't a turn end either. It sits above the reply, and only on turns where the model thought. Because the same line is also `message_start`, it counts as structural and the marker never anchors. `FullTurnText` on a thinking turn returns an empty string, or the startup banner when that is still on screen.

Dropping `turn_end` and `message_start` (gemini's shape) fixes all of it on the same frames. Busy frames read `working` with `TypingHold = "working"`, and the reply text comes back.

**2. Dialogs that read idle**

- Trust prompt. agy asks "Do you trust the contents of this project?" in every new directory, including a git worktree of a repo that is already trusted. That's every worktree spawn. It reads `idle` with `TypingHold = ""`, so a queued message would be typed into it.
- File-access prompt ("Allow access to this file?" with `> 1. Yes, allow access`). Also `idle` with the current rules, since only "Run this command?" is covered.
- `^>\s*\d+\.\s` also matches the echo of a prompt that opens with a numbered item (`> 1. reply with ...`) once `turn_end` is gone, and that session sticks on waiting.

All of these dialogs and the `/resume` picker draw `↑/↓ Navigate` in their footer. The resting composer never does, it shows `? for shortcuts`. This pair read every frame I captured correctly (busy, finished, thinking, permission, file access, trust, picker, numbered echo):

```toml
rules = [
  { state = "waiting", pattern = "↑/↓ Navigate" },
  { state = "working", pattern = "esc to cancel" },
]
```

**3. `agy -i /resume` sends /resume to the model**

On 1.2.9, `-i /resume` doesn't open the picker. The model gets "/resume" as a prompt, reasons about it and asks to run a tool (it tried `ls` once and a file read the second time). agy updates itself in the background, so users end up on this build. Until an id gets captured, every agy revive takes this path.

Typing `/resume` at the composer does open the picker, so opencode's shape fits: `resume_picker_command = "agy"` plus `resume_picker_keys = "/resume"`. That needs an `input_prefix` matching the bare `>` row, since `InjectPickerKeys` waits on it. One catch to check. `InjectPickerKeys` sends a second Enter 500ms later, and in agy's picker that selects the top row, which is the most recent conversation from any directory.

**4. Smaller things**

- A missing agy shows `install it with: brew install agy`, and no such formula exists. `deps.official` needs `"agy": "curl -fsSL https://antigravity.google/cli/install.sh | bash"`.
- No MCP registration yet. agy 1.2.9 has `agy mcp add <name> <command> [args...]`, so a `mcpreg` style like gemini's looks doable. Fine as a follow-up.
- `toolLabel` in `internal/report/report.go`, the README and `docs/configuration.md` / `docs/usage.md` don't list Antigravity yet. No tests cover the block. Real-pane frames in `status_test.go` would lock the fixes above in.
- `chrome_line` misses the logo rows that carry text, so a fresh session's row quotes `▄▀▀      ▀▀▄`.
- `limit_line` is still the unverified gemini string, as your comment says.
- Nit. The comment says the dialog rule must come before the busy match. `matchRules` already promotes a later waiting rule over a working one, so the order doesn't decide it.

On the restart test, what I said before stands, it's fine as a follow-up. Point 3 is why I'd want the picker path working in this PR, since until the capturer lands it's the only way back into a specific conversation.

PR #528, opened 2026-09-17
feat(sysstat): add a battery meter to the computer stats panel

zavatskiy:
## What changed

Added a `batt` line to the computer stats panel (TUI rail), directly under `disk`, reusing the existing meter/gauge rendering used for cpu/mem/swap/disk. Shows charge percent and, when charging, a "charging" suffix. Hidden entirely on machines with no battery or when the read fails.

## Why

The computer stats panel had cpu/mem/swap/disk/temp/net but no battery reading, even though the underlying data is readily available cross-platform via `github.com/distatus/battery`.

Closes #

### Required behavior

- Poll battery charge percent + charging state on the same interval as the other stats.
- Render with the existing `meter`/`gauge` helper — same bar style, same label width.
- Show a "charging" suffix when the state is `battery.Charging`.
- No battery present (`GetAll` returns an empty slice, or a fatal error like `ErrNotFound`): hide the line, no placeholder, no zero.
- Read errors: ignored, `BatteryOK` stays false, poll loop unaffected.
- Multiple batteries: **summed** (current and full capacity are summed across all batteries, then divided), not averaged — so a depleted second battery pulls the reported charge down instead of being diluted by a healthy one.
- Gauge color is intentionally inverted relative to cpu/mem/disk: `gauge` takes an `invert` argument, and the battery meter passes `true`. For load-based metrics high is bad, but for battery low charge is bad, so low % renders alarming (red) and high % calm — the same ramp, mirrored.

### Why this approach

`sysstat.sampleBattery` follows the same `sample*(snap *Snapshot)` shape as `sampleSwap`/`sampleDisk`/`sampleTemps`. The actual battery source is a package-level `var batterySource = battery.GetAll`, swappable in tests, matching how the rest of the package structures OS calls for testability.

## How it was verified

- [x] `go vet ./...` passes
- [x] `go build ./...` passes
- [x] Ran it against real sessions — built the branch, launched the TUI on a MacBook (macOS) and checked the `computer` block live: `batt` renders under `disk` with the charging suffix while plugged in, and the gauge color follows charge (green at 96%, while `disk` at 93% renders red).
- [x] Holds for every supported agent CLI and platform, or the description names what it leaves out — battery reads are OS-specific inside `distatus/battery`; this PR was built and tested on macOS only, but the library covers Linux/Windows/BSD/Solaris via its own per-platform files, and the sysstat-side logic is platform-agnostic.

## Visual evidence

**Before:** the `computer` block ends at `disk` / `temp`; no battery line.

**After** (text capture of the live TUI, plain text — color is not visible here):

```
█ disk ━━━━━━━━━━  93% 30.4GB free
█ batt ━━━━━━━━━╸  96%
█ temp soc 41°C
```

Color in the same capture: `disk` 93% is red (alarm), `batt` 96% is green (calm). A screenshot was not taken.

## Summary by CodeRabbit

* **New Features**
  * Added battery monitoring to system status information.
  * Displayed battery percentage in the computer details view.
  * Indicated when a device is charging.
  * Combined readings when multiple batteries are present.
  * Added battery-specific gauge coloring, making low charge more prominent and full charge calmer.

* **Bug Fixes**
  * Gracefully handled unavailable, unreadable, or unsupported battery data without disrupting system status display.


YoanWai (changes requested):
Thanks for this. Reusing the gauge and mirroring the ramp for charge both read well. I ran the branch merged with main (gofmt, vet and the full suite are clean), live in the TUI on an M2, and the Linux build against fake `/sys/class/power_supply` trees in Docker. A few things to change before it lands.

**1. On Linux, a wireless mouse hides the laptop battery**

`battery.GetAll` returns a `battery.Errors` whenever any single entry comes back partial. HID peripherals register in `/sys/class/power_supply` with `type` Battery and only `capacity` or `capacity_level`, with no `energy_*` or `charge_*` files (see `drivers/hid/hid-input.c` and `drivers/hid/hid-logitech-hidpp.c`). So a Logitech mouse or a Bluetooth keyboard makes every read return an error. Since 5b92602, `sampleBattery` returns early on it and the row never shows on that laptop.

The Linux build of each commit against the same trees:

| `/sys/class/power_supply` | 63380ca | 5b92602 | per-entry fix below |
|---|---|---|---|
| `BAT0` only | 75% | 75% | 75% |
| `BAT0` + `hidpp_battery_0` | 75% | hidden | 75% |
| `BAT0` + `hid-<addr>-battery` | 75% | hidden | 75% |
| `hidpp_battery_0` only (desktop) | hidden | hidden | hidden |
| empty | hidden | hidden | hidden |

CodeRabbit's point still stands. A failed `Current` read next to a good `Full` would show 0%. `GetAll` returns one error per battery in the same order, so dropping only the entries whose charge failed covers both cases:

```go
func sampleBattery(snap *Snapshot) {
	batteries, err := batterySource()
	var perBattery battery.Errors
	if err != nil {
		readErrors, ok := err.(battery.Errors)
		if !ok {
			return
		}
		perBattery = readErrors
	}
	var readable []*battery.Battery
	for index, entry := range batteries {
		if index < len(perBattery) && !chargeReadable(perBattery[index]) {
			continue
		}
		readable = append(readable, entry)
	}
	applyBatteries(snap, readable)
}

func chargeReadable(err error) bool {
	if err == nil {
		return true
	}
	partial, ok := err.(battery.ErrPartial)
	return ok && partial.Current == nil && partial.Full == nil
}
```

With it, your eight tests pass along with two more I used to check it. `battery.Errors{nil, battery.ErrPartial{Current: ..., Full: ...}}` keeps the laptop at 75%, and `battery.Errors{battery.ErrPartial{Current: ...}}` hides the row.

<details>
<summary>The fake sysfs trees</summary>

```sh
mk(){ d=$1; shift; mkdir -p "$d"; while [ $# -gt 0 ]; do printf '%s\n' "$2" > "$d/$1"; shift 2; done; }
mk sys/BAT0 type Battery status Discharging energy_now 30000000 energy_full 40000000 \
  energy_full_design 50000000 power_now 8000000 voltage_now 11500000 voltage_min_design 11100000
mk sys/hidpp_battery_0 type Battery scope Device status Discharging voltage_now 3900000 capacity_level Normal
GOOS=linux GOARCH=arm64 CGO_ENABLED=0 go test -c -o probe ./internal/sysstat
docker run --rm -v "$PWD/probe:/probe:ro" -v "$PWD/sys:/sys/class/power_supply:ro" alpine /probe -test.run <a test that prints sampleBattery>
```

</details>

**2. macOS reads 1 to 2 points under the menu bar**

The library's darwin reader divides `AppleRawCurrentCapacity` by `AppleRawMaxCapacity`. The menu bar and `pmset -g batt` show `CurrentCapacity` over `MaxCapacity`, which on Apple Silicon is the percent over 100. Samples on the M2 during this review:

| raw ratio (what the row shows) | `pmset` / menu bar |
|---|---|
| 40.8% | 43% |
| 37.8% | 39% |
| 33.1% | 34% |

The library doesn't expose those two keys, so darwin needs its own small reader. That means `ioreg -n AppleSmartBattery -r -a` into `howett.net/plist` (already pulled in by this PR), taking `CurrentCapacity`, `MaxCapacity` and `IsCharging`. I could only check Apple Silicon here.

That reader can also cache the way `cachedTemps` does. `ioreg` takes about 15ms here (median of 20 runs), and it runs inside `refreshOnce` under `runMu` on every other poll.

**3. The full-screen layout has no battery reading**

`fullFootLine` in `internal/ui/notices.go` builds the one-line meters for the full-screen layout and never reads `BatteryOK`. On this branch with the layout set to full and the battery at 33%, the foot reads `cpu 84%  mem 83% …  disk 82% 84.6GB free  temp soc 81°C  net …`, with no `batt`. A `batt` reading after `disk`, with the same `charging` suffix, covers it. A case in `fulllayout_test.go` would lock it in.

**4. Docs and visual evidence**

The Stats section of `docs/usage.md` lists every computer gauge, and the full-screen paragraph there names the condensed meters. Both need the battery, including that it stays hidden when no battery is present. Once the above is in, please add before and after frames of both layouts, following the [capture guide](https://github.com/YoanWai/agent-manager/blob/main/.github/CONTRIBUTING.md#capturing-a-tui-frame).

zavatskiy:
Thanks for the review. Pushed 432f05c, rebased onto current main.

1. **Linux peripherals:** `sampleBattery` now drops only the entries whose `Current` or `Full` read failed (your `chargeReadable`), so a HID battery next to `BAT0` no longer hides the row. Tests added for `Errors{nil, ErrPartial{...}}` (kept at 75%) and `Errors{ErrPartial{Current: ...}}` (hidden).
2. **macOS:** a darwin reader takes `CurrentCapacity`, `MaxCapacity` and `IsCharging` from `ioreg -n AppleSmartBattery -r -a` via `howett.net/plist`, cached for 10s. It feeds the same `applyBatteries`. Checked on Apple Silicon only.
3. **Full-screen layout:** `fullFootLine` shows `batt N%` after `disk`, with the `charging` suffix, and nothing when there is no battery. `TestFullFootShowsBattery` covers it.
4. **Docs:** `docs/usage.md` lists the battery gauge in Stats and in the full-screen meters, including that it stays hidden without a battery.

Frames follow the capture guide (110x30, throwaway HOME, own tmux socket), before = main, after = this branch. The popup in the split frames is a tmux config error from my sandbox, unrelated. At 110 columns the full-screen foot trims from the right, so `batt` is cut with `…` and `temp` drops out; that is the existing truncation.

<details>
<summary>before, split layout</summary>

```
  A G E N T   M A N A G E R                                                                0 agents · active
🬹🬂🬂🬂🬂🬂🬂🬂🬂🬂🬂🬂🬂🬂🬂🬂🬂🬂🬂🬂🬂🬂🬂🬂🬂🬂🬂🬂🬂🬂🬂🬂🬂🬂🬨───────────────────────────────────────────────────────────────────────────
█   root          no agents yet   ▐  root   0 agents
█                                 ▐  dir    /Users/sanek/src/agent-manager-worktrees/batt          inherited
█                                 ▐───────────────────────────────────────────────────────────────────────────
█                                 ▐  agents
█                                 ▐  (none yet, press space to spawn one)
█                                 ▐
█        no sessions yet          ▐
█                                 ▐
█          n starts one           ▐
█                                 ▐
█                                 ▐
█                                 ▐
█                                 ▐
█                                 ▐
█                                 ▐
█─────────────────────────────────▐
█ computer                        ▐
█ cpu  ━━━━━━━━━━ 100%            ▐
█ mem  ━━━━━━━━╸━  86% 13.7GB/16… ▐
█ swap ━━━━━━━━━━  84% 5.9GB/7.0… ▐
█ disk ━━━━━━━━━╸  96% 20.5GB fr… ▐
█ temp soc 56°C                   ▐
█ net  ↓ 4.3KB/s  ↑ 6.1KB/s       ▐
█                                 ▐
▀▄▄▄▄▄▄▄▄▄▄▄▄▄▄▄▄▄▄▄▄▄▄▄▄▄▄▄▄▄▄▄▄▄▟───────────────────────────────────────────────────────────────────────────
   Group     ↵ fold  double click fold  ←/→ close / open  o editor  r rename  m move  x/X kill / all
             v/V revive / all  a/u archive / restore  d delete
  View       ↑/k ↓/j navigate · n new · T terminal · g group · / search · t archived · w attention …
```

</details>

<details>
<summary>after, split layout</summary>

```
  A G E N T   M A N A G E R                                                                0 agents · active
🬹🬂🬂🬂🬂🬂🬂🬂🬂🬂🬂🬂🬂🬂🬂🬂🬂🬂🬂🬂🬂🬂🬂🬂🬂🬂🬂🬂🬂🬂🬂🬂🬂🬂🬨───────────────────────────────────────────────────────────────────────────
█   root          no agents yet   ▐  root   0 agents
█                                 ▐  dir    /Users/sanek/src/agent-manager-worktrees/batt          inherited
█                                 ▐───────────────────────────────────────────────────────────────────────────
█                                 ▐  agents
█                                 ▐  (none yet, press space to spawn one)
█                                 ▐
█        no sessions yet          ▐
█                                 ▐
█          n starts one           ▐
█                                 ▐
█                                 ▐
█                                 ▐
█                                 ▐
█                                 ▐
█─────────────────────────────────▐
█ computer                        ▐
█ cpu  ━━━━━━━━━━ 100%            ▐
█ mem  ━━━━━━━━━━  84% 13.5GB/16… ▐
█ swap ━━━━━━━━━━  84% 5.9GB/7.0… ▐
█ disk ━━━━━━━━━╸  96% 20.0GB fr… ▐
█ batt ━━━━━╸━━━━  59%            ▐
█ temp soc 55°C                   ▐
█ net  ↓ 3.7KB/s  ↑ 4.8KB/s       ▐
█                                 ▐
▀▄▄▄▄▄▄▄▄▄▄▄▄▄▄▄▄▄▄▄▄▄▄▄▄▄▄▄▄▄▄▄▄▄▟───────────────────────────────────────────────────────────────────────────
   Group     ↵ fold  double click fold  ←/→ close / open  o editor  r rename  m move  x/X kill / all
             v/V revive / all  a/u archive / restore  d delete
  View       ↑/k ↓/j navigate · n new · T terminal · g group · / search · t archived · w attention …
```

</details>

<details>
<summary>before, full layout</summary>

```
  A G E N T   M A N A G E R                                                                0 agents · active
🬹🬂🬂🬂🬂🬂🬂🬂🬂🬂🬂🬂🬂🬂🬂🬂🬂🬂🬂🬂🬂🬂🬂🬂🬂🬂🬂🬂🬂🬂🬂🬂🬂🬂🬂🬂🬂🬂🬂🬂🬂🬂🬂🬂🬂🬂🬂🬂🬂🬂🬂🬂🬂🬂🬂🬂🬂🬂🬂🬂🬂🬂🬂🬂🬂🬂🬂🬂🬂🬂🬂🬂🬂🬂🬂🬂🬂🬂🬂🬂🬂🬂🬂🬂🬂🬂🬂🬂🬂🬂🬂🬂🬂🬂🬂🬂🬂🬂🬂🬂🬂🬂🬂🬂🬂🬂🬂🬂🬂🬨
█   root                                                                                       no agents yet
█
█
█
█
█
█
█
█
█
█                                               no sessions yet
█
█                                                n starts one
█
█
█
█
█
█
█
█
█
█─────────────────────────────────────────────────────────────────────────────────────────────────────────────
█ cpu 100%  mem 85% 13.6GB/16.0GB  swap 84% 5.9GB/7.0GB  disk 96% 20.1GB free  temp soc …  messages 2  M open
▀▄▄▄▄▄▄▄▄▄▄▄▄▄▄▄▄▄▄▄▄▄▄▄▄▄▄▄▄▄▄▄▄▄▄▄▄▄▄▄▄▄▄▄▄▄▄▄▄▄▄▄▄▄▄▄▄▄▄▄▄▄▄▄▄▄▄▄▄▄▄▄▄▄▄▄▄▄▄▄▄▄▄▄▄▄▄▄▄▄▄▄▄▄▄▄▄▄▄▄▄▄▄▄▄▄▄▄▄▟
   Group     ↵ fold  double click fold  ←/→ close / open  o editor  r rename  m move  x/X kill / all
             v/V revive / all  a/u archive / restore  d delete
  View       ↑/k ↓/j navigate · n new · T terminal · g group · / search · t archived · w attention …
```

</details>

<details>
<summary>after, full layout</summary>

```
  A G E N T   M A N A G E R                                                                0 agents · active
🬹🬂🬂🬂🬂🬂🬂🬂🬂🬂🬂🬂🬂🬂🬂🬂🬂🬂🬂🬂🬂🬂🬂🬂🬂🬂🬂🬂🬂🬂🬂🬂🬂🬂🬂🬂🬂🬂🬂🬂🬂🬂🬂🬂🬂🬂🬂🬂🬂🬂🬂🬂🬂🬂🬂🬂🬂🬂🬂🬂🬂🬂🬂🬂🬂🬂🬂🬂🬂🬂🬂🬂🬂🬂🬂🬂🬂🬂🬂🬂🬂🬂🬂🬂🬂🬂🬂🬂🬂🬂🬂🬂🬂🬂🬂🬂🬂🬂🬂🬂🬂🬂🬂🬂🬂🬂🬂🬂🬂🬨
█   root                                                                                       no agents yet
█
█
█
█
█
█
█
█
█
█                                               no sessions yet
█
█                                                n starts one
█
█
█
█
█
█
█
█
█
█─────────────────────────────────────────────────────────────────────────────────────────────────────────────
█ cpu 100%  mem 82% 13.2GB/16.0GB  swap 84% 5.9GB/7.0GB  disk 96% 19.5GB free  batt 58% …  messages 2  M open
▀▄▄▄▄▄▄▄▄▄▄▄▄▄▄▄▄▄▄▄▄▄▄▄▄▄▄▄▄▄▄▄▄▄▄▄▄▄▄▄▄▄▄▄▄▄▄▄▄▄▄▄▄▄▄▄▄▄▄▄▄▄▄▄▄▄▄▄▄▄▄▄▄▄▄▄▄▄▄▄▄▄▄▄▄▄▄▄▄▄▄▄▄▄▄▄▄▄▄▄▄▄▄▄▄▄▄▄▄▟
   Group     ↵ fold  double click fold  ←/→ close / open  o editor  r rename  m move  x/X kill / all
             v/V revive / all  a/u archive / restore  d delete
  View       ↑/k ↓/j navigate · n new · T terminal · g group · / search · t archived · w attention …
```

</details>

YoanWai (approved):

YoanWai:
Thanks, merged. Ships in the next release.

PR #531, opened 2026-09-18
feat(ui): add kanagawa wave and lotus themes

dmitrydoni:
## What changed

Two entries in `internal/ui/theme.go` — `kanagawa wave` (dark) and `kanagawa lotus` (light) — plus the palette list in `docs/usage.md` and `kanagawa lotus` in the two light-theme lists in `theme_test.go`.

Nothing else moves. Picker order keeps the existing shape: `kanagawa wave` closes the dark block after `monochrome`, `kanagawa lotus` closes the light block after `paper`.

## Why

Kanagawa is a common terminal palette, and every shipped theme reads as a seam against it. The manager repaints the terminal background to match the theme (OSC 11), which is exactly the feature that makes a missing palette visible rather than merely a preference.

Closes #522

### Required behavior

Two palettes that are recognisably Kanagawa, resolved from the upstream source rather than translated by eye, and clearing the contrast floors in `TestThemeTextContrast`.

### Why this approach

Per the maintainer's note on #522, the palettes come from `lua/kanagawa/themes.lua` upstream, not from the bundled chroma styles: chroma's are syntax roles, these are UI roles. `ui` supplies the surfaces and the text, `syn` and `diag` supply the accents and the status colours. One rule per token, applied to both variants:

| Token | Upstream role | wave | lotus |
|---|---|---|---|
| `Bg` | `ui.bg` | `#1f1f28` sumiInk3 | `#f2ecbc` lotusWhite3 |
| `Surface` | surfaces, one step | `#2a2a37` sumiInk4 (`bg_p1`) | `#dcd5ac` lotusWhite1 (`bg_m2`) |
| `Overlay` / `Border` | surfaces, two steps | `#363646` sumiInk5 (`bg_p2`) | `#d5cea3` lotusWhite0 (`bg_m3`) |
| `Bright` | full-strength fg | `#dcd7ba` fujiWhite (`ui.fg`) | `#43436c` lotusInk2 (`ui.fg_dim`) |
| `Text` | the other fg | `#c8c093` oldWhite (`ui.fg_dim`) | `#545464` lotusInk1 (`ui.fg`) |
| `Dim` | `ui.special` | `#938aa9` springViolet1 | `#766b90` lotusViolet2 |
| `Subtle` / `Idle` | `syn.comment` | `#727169` fujiGray | `#8a8980` lotusGray3 |
| `Accent` | `syn.fun` | `#7e9cd8` crystalBlue | `#4d699b` lotusBlue4 |
[...]

PR #548, opened 2026-09-19
feat(ui): switch theme follows OS to the theme's own counterpart

### LINKED ISSUE #533: theme follows OS: switch to the theme's own light/dark counterpart
### What are you trying to do

Follow-up from #522, where you suggested a separate issue for this.

**What happens now.** With *theme follows OS* on, `autoThemeName` (`internal/ui/model.go`) keeps the stored theme when it is already on the detected side. On a mismatch it always jumps to `solarized light` or `classic`. So catppuccin mocha on a light Mac becomes solarized light, not catppuccin latte. Anyone who picked a palette family loses it every time the OS flips.

**What I have in mind.** On a mismatch, switch to the stored theme's own counterpart on the other side, when it has one. Otherwise, fall back to today's behaviour. Six families already ship both sides:

| dark | light |
|---|---|
| solarized dark | solarized light |
| catppuccin mocha | catppuccin latte |
| tokyo night | tokyo night day |
| gruvbox dark | gruvbox light |
| rosé pine | rosé pine dawn |
| kanagawa wave | kanagawa lotus (#531) |

`classic`, `nord`, `dracula`, `monochrome` and `paper` have no declared counterpart, so they keep the current fallback. I'm not proposing new pairings for them.

One way to do it: add a `Counterpart string` field on `Theme`, set on both sides of each pair. `autoThemeName` tries it before the fallback. A test can check that every counterpart names an existing theme on the opposite side of `lightBackdrop()` and points back. The existing `themeauto_test.go` cases stay valid, because nord and paper are unpaired.

Happy to send the PR once #531 lands, if the shape suits you.

### Anything else

agent-manager 0.37.0 on macOS.

dmitrydoni:
## What changed

- `Theme` gains a `Counterpart` field. The six families that ship both sides set it on both halves: solarized, catppuccin, tokyo night, gruvbox, rosé pine, kanagawa.
- `autoThemeName` tries the stored theme's counterpart on a polarity mismatch, before the `classic` / `solarized light` fallback.
- `TestThemeCounterpartsPair` checks that every counterpart exists, sits on the other side of `lightBackdrop()`, and names the theme back.
- `TestAutoThemeName` gains four cases: counterpart flips in both directions, and a paired theme already on the right side stays.
- `docs/usage.md` lists the pairs under **theme follows OS**.

## Why

With *theme follows OS* on, catppuccin mocha on a light Mac became solarized light, not catppuccin latte. Anyone who picked a palette family lost it on every OS flip.

Closes #533

### Required behavior

- On a mismatch, a paired theme switches to its own counterpart.
- An unpaired theme (`classic`, `nord`, `dracula`, `monochrome`, `paper`) keeps today's fallback. No new pairings are invented for them.
- A theme already on the detected side stays. An undetectable scheme changes nothing, as before.

### Why this approach

A field on `Theme`, not a side table next to `autoThemeName`, as agreed in #533. The pairing lives with the palette, so adding a family is one edit in `theme.go`. The symmetry test stops a future theme from half-pairing: a one-sided or same-side counterpart fails it. I checked this by pointing `kanagawa lotus` at `nord`, which fails with `kanagawa wave: counterpart "kanagawa lotus" points back to "nord"`.

## How it was verified

- [x] `go vet ./...` passes
- [ ] `go test -race ./...` passes. All packages pass except `internal/ui`, where tmux end-to-end tests fail on this machine with `no server running` / `server exited unexpectedly`. The same kind of failure happens on unmodified `main`: 7 failures there, 4 here, different tests on each run, and each passes when run alone. All theme tests pass (`-run 'Theme|LightBackdrop|AutoThemeName'`). CI is the reference.
- [x] `go build ./...` passes
- [x] Holds for every supported agent CLI and platform. Theme resolution is shared code, independent of tool and OS.

## Visual evidence

**Not applicable because:** the change adds no new visual state. It only picks which existing palette auto-detect lands on, and every palette involved already renders as shown in #531 and earlier. The table-driven cases in `TestAutoThemeName` pin the before/after mapping.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

## Summary by CodeRabbit

* **New Features**
  * Automatic theme selection now switches to the matching light or dark variant within the same theme family when available.
  * Unpaired themes continue to use the appropriate default fallback.

* **Documentation**
  * Updated theme behavior documentation with examples of paired themes and fallback handling.

* **Tests**
  * Added coverage for theme pair relationships and automatic light/dark theme switching.


dmitrydoni:
@coderabbitai On the rationale-only comments nitpick: keeping these as they are. The two `model.go` doc comments already existed upstream, and this only updates them for the counterpart step. The `Counterpart` field comment matches every other field in `Theme`, each of which has one. The test comment gives the reason for the test: stopping a theme from being paired on one side only, which is the invariant #533 asked for.

YoanWai (approved):

PR #549, opened 2026-09-20
fix(ui): never quote the queued-message block as the agent's reply

atirna:
Finishes the second half of #447: the queued-message block no longer reads as agent output.

When Claude's `message_start` anchor has scrolled out of the captured pane, `LastMessage` used to fall back to the newest content line above the box. With the queued-message block sitting there, that line is the tail of the manager's own directive, so the reply cell quoted the manager back to the user.

- `LastMessage` now yields an empty quote for tools that declare `message_start` and have no anchor on screen. The `anchored=false` signal already drives the transcript and deep-capture recoveries in the poller, so a real reply is recovered from Claude's transcript file or a deeper pane capture instead of guessed from unrelated rows. Tools without a marker keep the bare content-line fallback.
- The `❯ Press up to edit queued messages` row is claude chrome, so it is stepped over like the rest of the frame.

**This changes the unanchored fallback:** a marker tool's bare content line is no longer shown as the quote. That fallback was the bug: the content line above the box belongs to the queued block or our own directive, not the agent.

### Verification

- `go test ./internal/status/ ./internal/config/`: green, including new cases in `TestLastUserEchoAndScrolledMarker` (scrolled anchor → empty; queued block with `Press up to edit queued messages` → empty) and the unchanged anchored/`opencode` fallback cases.

Fixes the remaining half of #447

## Summary by CodeRabbit

- **Bug Fixes**
  - Improved pane status detection for tools that use message markers, preventing scrolled or incomplete content from being mistaken for the latest message.
  - Queued-message prompts are now excluded from pane content, improving composer and message-state recognition.
  - Empty or unavailable message content is reported accurately instead of displaying stale trailing text.


atirna:
checked this against the loader: no migration is needed here. LoadDir replaces cfg.Tools with the builtin table on every start (internal/config/config.go:194), and the starter config.toml a first run writes carries no [tools] section at all (config.go:295). stored files never held a chrome_line, and even one hand-added is ignored by design. the comment above builtinTools says exactly this: "builtinTools is the only source of tool definitions, so a release that fixes a CLI's new screen fixes it for everyone on that release" (config.go:322)

internal/status/status.go
```diff
@@ -403,6 +403,9 @@ func (e *Engine) LastMessage(tool, pane string) (line string, anchored, ok bool)
 		return "", false, true
 	}
 	if start == -1 {
+		if tr.messageStart != nil {
+			return "", false, true
```
YoanWai: **[bug]** An empty quote here drops the question a waiting session is asking. Claude blinks the `⏺` of a tool call while it runs, and an open `AskUserQuestion` dialog counts as one. In run 3 (nothing queued), 20 of the 41 dialog frames had the bullet off:

```
  Cat or dog?
────────────
 ☐ Pet pref

Cat or dog?

❯ 1. Cat
```

The base quotes `Cat or dog?` on all 41 frames. This branch returns `""` on the 20 bullet-less ones. The transcript recovery cannot fill it either. While the dialog is open the transcript has no assistant text (run 2's holds a `thinking` and a `tool_use` entry, run 3's no assistant entry yet), so `ClaudeTranscriptTail` returns an empty reply. The reply cell loses the question on every poll that lands on a bullet-less frame.

Blanking does clean up run 1, where the base quoted `owner`, the wrapped tail of a `⎿ Tip:` block. Runs 2 and 3 showed `◐ medium · /effort` and a tmux mouse hint the same way. Those rows are Claude chrome (a tip, the effort badge, a tmux hint), so they need their own handling rather than an empty quote on every bullet-less frame.

If the empty return stays, the doc comment on `LastMessage` still says `false` means the quote is the newest content line.

internal/status/status.go
```diff
@@ -403,6 +403,9 @@ func (e *Engine) LastMessage(tool, pane string) (line string, anchored, ok bool)
...
+		}
 		return strings.TrimSpace(lines[lastContent]), false, true
 	}
 	// The message runs from its marker until the next structural line: a
```
YoanWai: **[bug]** The queued message still reaches the quote whenever the bullet is drawn. Claude Code v2.1.281 draws it at column 0 as `❯ <text>` between the reply and the spinner or dialog, followed by `  ctrl+x ctrl+s to send now` while the turn runs. Neither row is structural, so this loop appends them. The base does the same, so this is the part of #447's second fix still open.

Two live quotes:

```
run 1: "Running python3 -c ... (ctrl+b ctrl+b (twice) to run in background) ❯ Run this exact shell command once, ... Then continue. ctrl+x ctrl+s to send now"
run 2: "Red or blue? ❯ Also tell me a fun fact about tmux after that."
```

In run 2 the queued text sits right after the question, so any row width shows it. An anchored quote skips the transcript recovery in `rowLines`, so it reaches the row unchanged.

**Suggestion:** stop the message at a prompt row. On the base, with the content-line fallback kept, breaking this loop on a `user_echo` match cleared the queued text from all 50 affected frames, kept `Red or blue?` and `Cat or dog?` on every dialog frame, and left `internal/status` and `internal/config` green. Gemini's `user_echo` is `^\s*> `, which also matches an indented `>` row, so check Gemini before applying the stop to every tool.

internal/config/config.go
```diff
@@ -353,7 +353,7 @@ status_source = "claude-hooks"
 default_status = "idle"
 activity_cutoff = "(?m)^❯"
 turn_end = "^[✻✳✶✽✢·✦✧+*] \\S+ for \\d.*$"
-chrome_line = "^\\s*[─q]{4,}.*$|^[\\s─q]*$|^\\s*✔ Update installed · Restart to update\\s*$|^\\s*new task\\? /clear to save .*$"
+chrome_line = "^\\s*[─q]{4,}.*$|^[\\s─q]*$|^\\s*✔ Update installed · Restart to update\\s*$|^\\s*new task\\? /clear to save .*$|^❯ Press up to edit queued messages\\s*$"
```
YoanWai: **[suggestion]** `❯ Press up to edit queued messages` is Claude's composer while messages are queued. It is the last `^❯` match, so it sits below the activity region. In all 58 captured frames that show it, it is outside the region, and the v2.1.261 capture in #447 draws it the same way. `TestLastUserEchoAndScrolledMarker` still passes with this alternative removed. Suggest dropping it.

The place this row does get read is `InputDraft`, which returns `"Press up to edit queued messages"` as a typed draft on all 58 frames. `input_placeholder` is the field that tells a placeholder from a draft, if you want Claude to learn this row.

internal/status/status_test.go
```diff
@@ -992,8 +992,12 @@ func TestLastUserEchoAndScrolledMarker(t *testing.T) {
 	if !ok || anchored {
 		t.Fatalf("scrolled pane: anchored=%v ok=%v, want unanchored", anchored, ok)
 	}
-	if line != "tail of a reply that scrolled its bullet away." {
-		t.Fatalf("scrolled fallback = %q", line)
+	if line != "" {
+		t.Fatalf("scrolled fallback = %q, want empty for a marker tool", line)
+	}
+	if line, _, ok := engine.LastMessage("claude",
+		"  directive the manager typed\n  more of the same block\n❯ Press up to edit queued messages\n❯ "); !ok || line != "" {
```
YoanWai: **[suggestion]** Claude draws no second `❯ ` composer under `Press up to edit queued messages`, and v2.1.281 does not indent the queued text. This pane passes on the no-anchor rule alone. A fixture trimmed from a real frame, bullet on, would lock the bug on `status.go:411`. This one returns `Red or blue? ❯ Also tell me a fun fact about tmux after that.` on this branch and on the base, and `Red or blue?` with the stop at `user_echo`:

```
⏺ Red or blue?

❯ Also tell me a fun fact about tmux after that.
────────────
 ☐ Color

Red or blue?

❯ 1. Red
     Red
  2. Blue
```

The bullet-less frame on line 407 makes a good second case, expecting `Cat or dog?`.

YoanWai (changes requested):
## Summary

Thanks for picking up the second half of #447. I replayed three live Claude Code v2.1.281 sessions through `status.Engine` at `68014d6` and at its base `77914fd`, and two findings change the picture.

Claude blinks the `⏺` of a tool call that is still running, and an open `AskUserQuestion` dialog is one. So a missing bullet usually means the blink is off, and the message is still on screen. On those frames the base quotes the question the session is waiting on, and this branch quotes nothing (line 407).

The queued message does still reach the reply cell, on the frames where the bullet is drawn (line 411). On bullet-less frames the base never picked it up. Across 43 such frames with a queued message on screen, current Claude drew it above the spinner or the dialog, and the base's fallback landed below it.

Your reply on the migration thread is right. `LoadDir` replaces the stored tools with the built-in table, so a `chrome_line` edit reaches every install.

"Fixes the remaining half of #447" does not link the issue (`closingIssuesReferences` is empty), so the body needs a `Fixes #447` line if this should close it.

## How I checked

Each session ran on an isolated tmux socket and was captured every 0.5s with `tmux capture-pane -p`. Every frame went through `LastMessage`, `ActivityRegion` and `InputDraft` with `config.Default()`.

| Run | Frames | Base `77914fd` | This PR |
|---|---|---|---|
| 1. 30s foreground Bash call, message queued 5s in, bullet off | 38 | `"owner"` (tail of a wrapped Claude tip) | `""` |
| 1. same, bullet on, queued message on screen | 29 | queued text in the quote | queued text in the quote |
| 2. `AskUserQuestion` with a message queued, bullet off | 15 | `"Red or blue?"` | `""` |
| 2. same, bullet on | 21 | queued text in the quote | queued text in the quote |
| 3. `AskUserQuestion`, nothing queued, bullet off | 20 | `"Cat or dog?"` | `""` |
| 3. same, bullet on | 21 | `"Cat or dog?"` | `"Cat or dog?"` |

`go test ./...` is green apart from tmux lifecycle tests in `internal/ui` that fail on a different test each run and pass when rerun alone.

## Issue counts by severity

- bugs: 2
- suggestions: 2
- nits: 0

atirna:
pushed the rework. LastMessage is back to its base shape; the ❯ row a queued message or echoed prompt draws at the left edge is claude chrome now, so an anchored quote ends at it and the fallback steps over it, the ctrl+x hint, the effort badge and the tmux hints.

- took the stop through chrome_line per tool rather than a user_echo break in the loop: gemini's echo is `^\s*>` which also matches > rows inside a reply, so a tool-wide stop would cut gemini quotes
- added input_placeholder for the queued composer row, so InputDraft stops reading it as a draft
- fixtures are the frames from your replay now, incl the bullet-less dialog expecting Cat or dog?
- body has a real Fixes #447 line

go test ./internal/status/ ./internal/config/ green, gofmt/vet/build clean

YoanWai:
@coderabbitai full review

YoanWai:
@coderabbitai full review

YoanWai:
@coderabbitai full review

internal/status/status.go
```diff
@@ -412,19 +415,39 @@ func (e *Engine) LastMessage(tool, pane string) (line string, anchored, ok bool)
...
+	}
+	open := false
+	for i, raw := range lines {
+		line := strings.TrimRight(raw, " \t")
+		if strings.TrimSpace(line) == "" {
+			open = false
+			continue
+		}
+		open = open || tr.chromeBlock.MatchString(line)
+		inBlock[i] = open
```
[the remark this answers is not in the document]
YoanWai: Not taking this one. It fails the reachability step.

1. Reachable: no. Claude 2.1.281 separates every block with a blank row. Across 764 live pane captures (queued messages, open dialogs, running tool calls, finished turns) there are 0 frames with a `⏺` row directly under a `❯`, spinner or logo row. The flush `❯`/`⏺` shapes in the tests are hand-written `FullTurnText` fixtures, where `newestTurn` slices the echo off before `contentRows` runs, and they pass on this head.
2. Real: not reproducible from a real frame.
3. In scope: yes, `chromeBlockRows` is new in this branch.
4. Cost: about 3 lines, spent guarding an input Claude does not draw.

A capture of a Claude release that draws the reply flush under the echo would be the evidence to add the close.

internal/status/status.go
```diff
@@ -536,15 +559,19 @@ func (tr toolRules) turnProse(body []string, afterEcho bool) string {
 // trimPrompt drops indented rows until the first row at the left edge.
 func (tr toolRules) contentRows(body []string, trimPrompt bool) []string {
 	out := make([]string, 0, len(body))
+	inBlock := tr.chromeBlockRows(body)
```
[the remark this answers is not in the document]
YoanWai: Leaving this for a separate change. It fails the scope step.

1. Reachable: yes. Copying the reply while a queued message is on screen makes `lastEchoIndex` pick the queued `❯` row as the turn start.
2. Real: yes, and identical on main. Replaying the live captures, `FullTurnText` returns the same text on `d2c4fd4` and on this head for every queued frame, for example `Red or blue?` on the queued dialog.
3. In scope: no. The turn boundary comes from `newestTurn` and `lastEchoIndex`, which this branch does not touch. The branch only drops chrome-block rows inside the turn already chosen.
4. Cost: telling a queued echo from a submitted one needs its own rule in `lastEchoIndex` plus tests, well past 10 lines for a copy issue that is not a crash.

Noted as a follow-up.

YoanWai:
@coderabbitai full review

YoanWai (approved):
Approving with `aadf7a9` on top of your rework, @atirna. Thanks for taking the second half of #447.

Your commit fixed the queued text on frames where the bullet is drawn. On live Claude 2.1.281 the badge and tmux hints are right-aligned, and a queued message or spinner tip that wraps leaves its second row unmatched, so the reply cell still showed them in 14 of 70 captures of the running manager. `aadf7a9` adds a `chrome_block` rule: a queued message or prompt echo, the spinner and the welcome logo each own the rows drawn straight under them. That took it to 0 of 72, with status matching unchanged across 622 captured frames.

PR #551, opened 2026-09-20
fix(ui): keep hide empty out of the archived view

### LINKED ISSUE #484: HIDE EMPTY badge still paints over the archived list
### What happened

Turning on hide empty groups (`e`) and then switching to the archived view (`t`) still paints a `HIDE EMPTY` badge over the archive rail, next to `ARCHIVED`.

`filterBadgeLines` stacks every active narrowing without asking which view the list is in:

```go
if m.showArchived {
    badge("ARCHIVED", "t", "back to active")
}
if m.statusFilter.active() {
    badge(strings.ToUpper(m.statusFilter.label()), "w", "show all")
}
if m.hideEmptyGroups {
    badge("HIDE EMPTY", "e", "show empty")
}
```

The View legend does the same: `e` still reads `hide empty` / `show empty` while `t` already says `back to active`.

Hide empty is a filter on the live tree's unused group scaffolding. The archived view already prunes to archived sessions and archived groups, and it has its own `ARCHIVED` badge. The empty-groups label does not belong there. `TestFilterBadgesStackOverTheList` currently requires all three badges together, including `HIDE EMPTY` on an archived rail.

### What I expected

The archived list should show `ARCHIVED` (and `ATTENTION` when that filter is on). `HIDE EMPTY` should only appear on the active list.

### Steps

1. Press `e` so empty groups are hidden (`HIDE EMPTY` appears).
2. Press `t` to open the archived view.
3. The rail still shows `HIDE EMPTY` under `ARCHIVED`.

### agent-manager version

0.36.0

### Operating system

macOS

### tmux version

tmux 3.7b

### Which agent CLI

Grok Build

### Agent CLI version

grok 1.0.25 (f7e67d6988e2) [stable]

Michaelnogh:
## What changed

The archived view no longer shows the `HIDE EMPTY` badge, no longer offers the `e` hide/show empty key in the legend, and no longer applies the hide-empty filter to its rows.

## Why

Closes #484

### Required behavior

With hide empty groups on, switching to the archived view shows `ARCHIVED` (plus `ATTENTION` when that filter is on) and nothing about empty groups. The flag stays set, so the active list still hides empty groups when you go back.

### Why this approach

Hide empty is a filter on the live tree's unused group scaffolding. The archived view already prunes to archived sessions and groups, so the flag has nothing to say there. The guard is `!m.showArchived` at the three places that read the flag: `filterBadgeLines`, `rebuildRows` and the View legend. No new state.

## How it was verified

Drove a built binary against real tmux sessions on a throwaway socket and `HOME`, with an active session, an archived session and empty groups on both sides. Pressing `e` then `t` shows only `ARCHIVED` on the archived rail, the legend drops the `e` hint there, and archived empty groups stay listed. Pressing `t` back restores `HIDE EMPTY` and the active list still hides empty groups.

- [x] `go vet ./...` passes
- [x] `go build ./...` passes
- [x] Holds for every supported agent CLI and platform: the change is in shared list rendering, not tool specific

`TestFilterBadgesStackOverTheList` no longer expects `HIDE EMPTY` on the archived rail. New tests cover the badge (`TestHideEmptyBadgeBelongsToTheActiveRail`), the legend (`TestLegendOffersEmptyGroupsOnlyInTheActiveView`) and the rows (`TestHideEmptyGroupsDoesNotFilterTheArchivedView`).

## Visual evidence

**Not applicable because:** the change removes one badge and one legend hint from the archived view, and rendering tests assert both.


YoanWai (changes requested):
Please include before and after screenshots. The archived view should drop the HIDE EMPTY badge and the empty-groups key from the legend. The active view should still show both.

The [capture guide](https://github.com/YoanWai/agent-manager/blob/main/.github/CONTRIBUTING.md#capturing-a-tui-frame) shows how to capture a TUI frame.

Pressing e in the archived view still changes the hide-empty setting, even though the badge and key hint are gone. Press e, t, e, then t again. You return to the active list with hide empty turned off and no visible feedback while in archives. Please make e do nothing in the archived view and add a test for that sequence. I reproduced this with a failing test.

A small cleanup too. The new comments above the badge and legend tests repeat the test names. Please remove them and the added archive explanation in rebuildRows. The comment above toggleEmptyGroups also needs updating because it still says the filter applies to archives.

Michaelnogh:
All four addressed in 1e5ae1c.

**The `e` key in the archived view.** You were right, and the sequence you gave reproduces it exactly. The badge, the legend hint and the row filter were all guarded, but the key itself was not, so `e` flipped `hideEmptyGroups` with nothing on screen reporting it. `toggleEmptyGroups` now returns early when `showArchived` is set. I put the guard in the function rather than the `keybind.EmptyGroups` arm of `handleKey`, so the dispatch stays a plain call like its neighbours; say the word if you would rather see it in the switch.

**Test for the sequence.** `TestEmptyGroupsKeyIsRefusedInTheArchivedView` drives `e`, `t`, `e`, `t` through `handleKey` and asserts the active list is still filtered on return. I checked it fails without the guard: `e in the archived view should leave the hide-empty setting alone`.

**Comments.** Removed the two above the badge and legend tests, and the archive sentence in `rebuildRows`.

**`toggleEmptyGroups` doc.** It claimed the filter applied to the active and archive views both, which is what made the live key look deliberate. It now says what the filter covers and why the key is refused in archives.

**Frames.** In the PR body, as text captures rather than images: `main` and this branch, each driven on a throwaway socket and `HOME` at 150x34 from an empty store, per the capture guide. Steps 3 and 4 are the interesting pair. On `main` the legend flips to `e hide empty` while you are in the archives, and the empty groups are back on the active list afterwards. On this branch both are unchanged.

YoanWai (approved):
Thanks, all four points are in. I ran e, t, e, t live and the setting now survives the trip through the archives.

YoanWai:
Thanks @Michaelnogh. The before and after frames made this easy to review, and so did the test for the e, t, e, t sequence. It's merged and ships in the next release.

